PR #3, opened 2017-07-10
set up mkdocs

devinreams:
if we do this then we can set up the repo to use GitHub Pages, and then `mkdocs gh-deploy` to deploy to the `gh-pages` branch


devinreams:
Done! and live at:

https://mozilla-lockbox.github.io/lockbox-extension/

PR #7, opened 2017-07-11
some requires added after upgrading to npm 5.2.0

devinreams:
after running `npm install` I got some changes to my `package-lock.json`

if you check this out does it still work OK for you, @linuxwolf?


devinreams:
We can back this out if needed.

Going to merge so future npm stuff is clean and clear.

PR #10, opened 2017-07-12
Support React

jimporter:
This PR adds support for React and building via WebPack. We should probably decide what (if anything) to do with the Makefile, since there's now a build step that needs to run. Also, I'd really like it if we had a better way of managing non-WebPacked files (i.e. the stuff in `dist/`), since I hate having to cram the build artifacts in there and add them to .gitignore. I'd much prefer out-of-tree builds.


package.json
```diff
@@ -13,16 +13,30 @@
...
   "keywords": [
     "passwords",
     "crypto",
     "security",
     "addon"
   ],
+  "scripts": {
+    "run": "jpm run",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
```
linuxwolf: There should be an entry for running webpack here, e.g.:

```
  "build": "webpack"
```

devinreams:
For testing out our Project board: I believe this fixes #8 
(which would now auto-close when this merges)

linuxwolf:
1. My preference is to replace the Makefile with npm run scripts.
2. Is there a reason not to add all of dist/ to `.gitignore`?  I realize it means extra copying, but it keeps the distribution consolidated.  This is where `watchman` helps.

.gitignore
```diff
@@ -1,3 +1,4 @@
-node_modules/
+/node_modules/
+/site/
 /lockbox-*.xpi
-/site
+/dist/chrome/content/aboutPage.js
```
linuxwolf: why not all of `/dist/`?

dist/chrome/content/aboutPage.html
```diff
@@ -5,8 +5,9 @@
 <html>
 <head>
   <title>Lockbox</title>
+  <script src="chrome://lockbox/content/aboutPage.js" defer></script>
```
devinreams: Now that this is back to about:lockbox, is this path correct?

devinreams: Follow-up: yes, it is. I just need to run webpack to actually get this asset in the right place.

devinreams:
worksonmymachine

PR #21, opened 2017-07-18
Put the management page inside the embedded WebExtension

jimporter:
This unfortunately means that we can't have about: URLs for now, so the
nsIAboutModule implementation is gone.


devinreams:
Works here! ✅ 

For posterity, this is the URL the browser action loaded

`moz-extension://2a7ba4f5-d7e0-d142-b95b-ff7ad8789b30/manage.html`

Dat GUID. 😋

jimporter:
Just a quick note, the UUID changes every time (I think it's per-install) so we can't predict what a given user's URL will look like.

PR #33, opened 2017-07-27
build: fixes/enhancements to npm build scripts

linuxwolf:
This does a couple of things:

* enhances `package` to generate the name from the `package.json` fields
* fixes `lint` to cover all of the JavaScript sources
* fixes `clean` to not delete all of `/node_modules` (otherwise `package` can fail)
* added `distclean` to get rid of **EVERYTHING** not trackable

This does not address issues building on Windows, which is a little more involved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,13 +25,15 @@
   ],
   "scripts": {
     "prerun": "npm run-script build",
-    "run": "jpm run --addon-dir dist/ $npm_config_jpm_runflags",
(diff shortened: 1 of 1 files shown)


linuxwolf:
Not typically no.  It's not necessary, but I've seen a lot at least put `pre` trigger scripts just before (and `post` trigger scripts just after) their targets to help visualize the flow.

PR #34, opened 2017-07-31
Don't copy extra JS files to the dist dir

jimporter:
This switches us to opt-in for copying to dist. That way, we don't
accidentally copy over JS files that are already bundled by WebPack (this was
happening for l10n.js).

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,44 +1,11 @@
-var CopyWebpackPlugin = require("copy-webpack-plugin");
+const CopyWebpackPlugin = require("copy-webpack-plugin");
+
 var path = require("path");
 
 var entries = {
   background: "./webextension/background.js",
-  manage: ["babel-polyfill", "./webextension/manage.js"]
+  manage: "./webextension/manage.js"
 };
-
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -56,16 +23,22 @@ var config = {
...
   plugins: [
     new CopyWebpackPlugin([
-      {from: "**/*"},
-    ], {ignore: ignoreOnCopy}),
+      {from: "bootstrap.js"},
+      {from: "chrome.manifest"},
+      {from: "install.rdf"},
+      {from: "webextension/manifest.json", to: "webextension/"},
+      {from: "webextension/en-US.ftl", to: "webextension/"},
+      {from: "webextension/manage.html", to: "webextension/"},
```
linuxwolf: would it make sense to change this to "webextension/*.html"?

I'm worried about inadvertently forgetting files.

jimporter:
Yeah, it's hard to balance "including only the files we want" and "not missing any files we want". I switched a few lines to globs so it should hopefully be a little better though...

PR #35, opened 2017-08-04
Support Redux; resolves #24

### LINKED ISSUE #24: Integrate Redux into extension
This may all go together as part of getting something storing/retrieving data at https://github.com/mozilla-lockbox/lockbox-extension/issues/17 but calling it out separate in case it's helpful

jimporter:
Still not 100% sure about this, but it's a start. I'm still looking into whether it would make more sense to have the Redux store live in the background scripts.

resolves #24


devinreams:
Works on my machine! 🌟 

lockbox-redux

PR #45, opened 2017-08-29
Add Mocha tests

jimporter:
Ok, everything here should be working. Some useful tidbits:

* The test just makes sure we can mount the `App` component successfully; this entails setting up a mock Redux store and fake l10n.
* I added source maps so it'll be a lot easier to tell where things go wrong (the stack traces no longer point to the bundled JS files).
* I removed the `.babelrc` file for now, though we should probably add it back later. As it was, it broke running the app because the configuration was conflicting with the real babel config in `webpack.config.js`. Later, we can put everything in `.babelrc` and use its [NODE_ENV](https://babeljs.io/docs/usage/babelrc/#env-option) support, but that seems logically separate from this.


test/manage-test.js
```diff
@@ -0,0 +1,37 @@
...
+const middlewares = [];
+const mockStore = configureStore(middlewares);
+
+// Keep this in sync with <src/webextension/manage/reducers.js>.
+const initialState = {
+  storage: {entries: []},
+  ui: {selectedEntry: null, newEntry: false},
+};
+
+describe("It renders without crashing", () => {
```
m8ttyB: \o/

test/mock-l10n.js
```diff
@@ -0,0 +1,18 @@
...
+function *generateMessages() {
+  const context = new MessageContext("en-US");
+  yield context;
+}
+
+export default function MockLocalizationProvider(props) {
+  return (
+    <LocalizationProvider messages={generateMessages()} {...props}/>
+  );
+}
```
m8ttyB: Interesting 👍

webpack.config.test.js
```diff
@@ -0,0 +1,38 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const combineLoaders = require("webpack-combine-loaders");
+const nodeExternals = require("webpack-node-externals");
+const path = require("path");
+
+const config = {
+  target: "node",
+  externals: [nodeExternals()],
+  devtool: "cheap-module-source-map",
```
m8ttyB: +1

package.json
```diff
@@ -26,7 +26,7 @@
   "scripts": {
     "prerun": "npm run-script build",
     "run": "jpm run --addon-dir dist/ ${npm_config_jpm_runflags} --prefs jpm-prefs.json",
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "mocha-webpack --opts ./mocha-webpack.opts --webpack-config ./webpack.config.test.js -r test/setup.js test",
```
linuxwolf: It's generally recommended to have a "pretest" script that ensures linting is run:

```
  "pretest": "npm run lint"
```

.travis.yml
```diff
@@ -7,3 +7,4 @@ before_script:
 
 script:
   - npm run lint
+  - npm test
```
linuxwolf: Note that, by default, Travis does the following:

```
before_install:
  - npm install

script:
  - npm test
```

Linting should run every time tests are run (see comments on `package.json`).

webpack.config.test.js
linuxwolf: I keep forgetting webpack expanded the number of devtool options ...

linuxwolf:
Also, runs on my machine! 👍

package.json
jimporter: I'd actually prefer to go the other direction and turn off auto-linting for the `build` command too. It's pretty frustrating when I'm just slapping code together for debugging and I have to fix linter issues before I can actually test my code. If CI is always linting, we should never run into lint failures accumulating...

As a compromise, I guess we could add a `NO_LINT` environment variable, but I'm a little worried that that would make the resulting commands POSIX-only (I sort of doubt that the Windows shell syntax is identical).

linuxwolf: [`cross-env`](https://github.com/kentcdodds/cross-env#readme) takes care of differences between POSIX and Windows.

I understand not wanting the lint to get in the way, but my experience has been it has fixed more problems than it causes -- even in troubleshooting mode.  Especially with `no-console` set to "warn" as it is here.

jimporter: My main issue is that the linter tries to address two classes of problems: stuff that would get caught by `-Wall` and true lint issues (e.g. requiring a space after a comment token). The latter problems are usually just noise that interrupt me during debugging, whereas the former are actually useful.

Maybe running lint as a pre-commit hook would be a good compromise? Then every commit should be fully-linted.

linuxwolf: Hooks in github would be better, although I'd still want linting run as part of Travis (so maybe I don't trust all the hooks (-: ).

The current setup is ok, and I'd rather have this than wait for "perfect".

jimporter:
I was thinking of doing pre-commit hooks because I wanted to have an easy way to make sure I've linted locally before I push anywhere. Doing it for each commit seems like a good compromise for me, since then it's easy to do sloppy stuff during debugging, but you have to clean it up before you commit. That plus Travis runs on PRs should catch everything we need.

I'll file a followup issue for us to decide how we want to handle this.

devinreams:
yes

PR #44, opened 2017-08-24
Hook up to the real datastore

jimporter:
This is still WIP and isn't hooked up to the real datastore yet, but I made a fake datastore in the background script, plus added local sync logic to the management UI so that the management page is never stale.

when merged, this fixes #30

PR #46, opened 2017-08-29
Add some real unit tests

### LINKED ISSUE #28: Write unit tests that run locally with: mocha, chai, assert
Things we need to test:

* [ ] `bootstrap.js`
* `webextension/`
  * [x] `common.js`
  * `background/`
    * [x] `datastore.js`
    * [ ] `index.js`
    * [x] `messagePorts.js`
  * `manage/`
    * `components/`
      * [x] `app.js`
      * [x] `item.js`
      * [x] `itemDetails.js`
      * [x] `itemList.js`
    * `containers/`
      * [x] `addItem.js`
      * [x] `allItems.js`
      * [x] `currentItem.js`
    * [x] `actions.js`
    * [ ] `index.js`
    * [ ] `l10n.js`
    * [x] `reducers.js`

jimporter:
This PR is WIP and based on #44. I'll try to add a few more useful tests and then put this up for review.

Resolves #28


jimporter:
This is ready for review now. Note: I didn't write tests for <CurrentItem/>, but that's because I plan to change it quite a bit next, and it didn't seem worthwhile to test something that'll have a much-different API. I also haven't tested the actions or reducers yet; those are coming later.

test/manage/components/app-test.js
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+require("babel-polyfill");
+
+import chai, { expect } from "chai";
```
linuxwolf: When last we talked about testing, we agreed on `assert` not `expect`.

It's not really a problem, but it's a surprise that would have been nice to get at least a notice about.

Note `lockbox-datastore` is using `assert`.

package.json
```diff
@@ -26,7 +26,8 @@
   "scripts": {
     "prerun": "npm run-script build",
     "run": "jpm run --addon-dir dist/ ${npm_config_jpm_runflags} --prefs jpm-prefs.json",
-    "test": "mocha-webpack --opts ./mocha-webpack.opts --webpack-config ./webpack.config.test.js -r test/setup.js test",
+    "test": "cross-env NODE_ENV=test mocha-webpack --opts ./mocha-webpack.opts --webpack-config ./webpack.config.test.js -r test/setup.js test",
+    "coverage": "nyc --all --include \"src/**/*.js\" npm test",
```
linuxwolf: I really don't agree with this breakout.

I sort of understand the concern about performance,but only sort of.

test/manage/components/app-test.js
jimporter: I think I just used `expect` because it's what @m8ttyB's original patch started with (and I forgot what we'd agreed on). However, `expect` *does* give us more options in terms of plugins so maybe it's the better choice long-term...

package.json
jimporter: I don't have a strong opinion here, but I split them out just so that it's easier to see the test results during dev.

jimporter: I re-merged `coverage` and `test` since it'll probably change anyway when we switch to using karma.

linuxwolf: 👍

PR #47, opened 2017-09-01
Provide production configs

### LINKED ISSUE #40: Support production build config
Eventually, we should ensure that everything we do in this add-on is production-ready (e.g. minifying JS). I'll try to keep an up-to-date list of all the stuff we need:

* [ ] Minified JS
* [ ] [Compiled CSS](https://javascriptplayground.com/blog/2016/07/css-modules-webpack-react/)

jimporter:
This resolves #40. It minifies JS, compiles all our CSS into one file, and even minifies our (tiny) HTML file!


.eslintrc.js
```diff
@@ -1,4 +1,8 @@
-{
+let extraPlugins = [];
+if (process.env.STRICT_LINT !== "1")
```
linuxwolf: needs curly braces.

webpack.config.babel.js
```diff
@@ -52,13 +92,21 @@ export default {
...
-      {from: "webextension/**/*.html"},
       {from: "webextension/icons/*"},
     ]),
     new webpack.DefinePlugin({
       "process.env": {
         "NODE_ENV": JSON.stringify(NODE_ENV),
       },
     }),
+    new HtmlWebpackPlugin({
+      title: "Lockbox",
```
linuxwolf: This string is in a bunch of places.  At some point, we really ought to figure out how to parameterize it project-wide.

*eventually*

.eslintrc.js
jimporter: Fixed, and I also moved this to #46, since it really belongs in that PR. I also set eslint to lint this file so it can't happen again.

PR #50, opened 2017-09-01
import all exported DataStore symbols

linuxwolf:
This change works the same while `lockbox-datastore` follows CommonJS or if (when?) it switches to ES6 modules.


jimporter:
@linuxwolf What does this do? Do we need to explicitly import all the exported symbols? Maybe this just depends on what you're thinking about for `lockbox-datastore`. (Or my lack of understanding about the nuances of `import`...)

linuxwolf:
This is future-proofing.  There are some nuanced differences between CommonJS modules and ES6 modules that this syntax papers over.

PR #54, opened 2017-09-05
Add some basic styles

devinreams:
@jimporter I wanted to quickly play around with CSS Grid and make some quick UI changes.

## Before

screen shot 2017-09-05 at 12 31 54 pm

## After

screen shot 2017-09-05 at 12 29 39 pm

Are you okay with these style updates?

### DIFF
--- a/src/template.ejs
+++ b/src/template.ejs
@@ -13,6 +13,7 @@
     <% } for (let js of htmlWebpackPlugin.files.js) { %>
     <script src="<%= path.relative(dir, js) %>" defer></script>
     <% } %>
+    <link href="<%= htmlWebpackPlugin.options.icon %>" rel="shortcut icon" type="image/png"/>
   </head>
   <body>
     <main id="content"></main>
--- a/src/template.ejs
+++ b/src/template.ejs
@@ -14,6 +14,8 @@
     <script src="<%= path.relative(dir, js) %>" defer></script>
     <% } %>
     <link href="<%= htmlWebpackPlugin.options.icon %>" rel="shortcut icon" type="image/png"/>
+    <link href="chrome://browser/content/extension.css" rel="stylesheet">
+    <link href="chrome://browser/content/extension-mac.css" rel="stylesheet">
   </head>
   <body>
     <main id="content"></main>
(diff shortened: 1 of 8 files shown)


jimporter:
I think we want to use the built-in browser styles if at all possible. I've tried to get this working, but no luck as of yet...

devinreams:
>I think we want to use the built-in browser styles if at all possible.

@jimporter Ah right, good call...

Based on what I've read, I think we have to link the stylesheets in to the page ourselves (and then add `browser-style` class to some elements). I may not have done it "properly" but took a quick stab with my latest commits.

For my own future reference: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/user_interface/Browser_styles

src/template.ejs
```diff
@@ -13,6 +13,9 @@
     <% } for (let js of htmlWebpackPlugin.files.js) { %>
     <script src="<%= path.relative(dir, js) %>" defer></script>
     <% } %>
+    <link href="<%= htmlWebpackPlugin.options.icon %>" rel="shortcut icon" type="image/png"/>
+    <link href="chrome://browser/content/extension.css" rel="stylesheet">
```
devinreams: This should definitely be happening automatically, but just threw it in there for now to test

src/webextension/manage/containers/addItem.js
```diff
@@ -12,7 +12,8 @@ import { startNewItem } from "../actions";
 function AddItem({dispatch}) {
   return (
     <Localized id="add-item">
-      <button onClick={() => { dispatch(startNewItem()); }}>
+      <button className="browser-style"
+        onClick={() => { dispatch(startNewItem()); }}>
```
jimporter: Nit: indent this so `className` and `onClick` line up vertically.

src/template.ejs
jimporter: I'm not actually sure it should be automatic, since we're just opening a page via the WebExtension. There's nowhere I know of to set `browser_style` the way we're doing things. It might make sense to have this be optional though (e.g. pass `browser_style` to the *template*).

devinreams: >I'm not actually sure it should be automatic, since we're just opening a page via the WebExtension

Ah yeah, its usually just for popups and options pages I think? MDN got me a little confused.

For my own reference again: https://dxr.mozilla.org/mozilla-central/rev/955840bfd3c20eb24dd5a01be27bdc55c489a285/browser/components/extensions/ext-utils.js#62

Needs more rethinking but low priority, will come back later.

jimporter:
Let's land this. We can do more styling later as well.

PR #57, opened 2017-09-06
Add Coveralls (test coverage) to CI and readme

devinreams:
I mostly wanted the fancy badge...

screen shot 2017-09-05 at 10 09 31 pm

screen shot 2017-09-05 at 10 11 19 pm

What do you think @jimporter @m8ttyB?


package.json
```diff
@@ -31,6 +31,7 @@
     "run": "jpm run --addon-dir dist/ ${npm_config_jpm_runflags} --prefs jpm-prefs.json",
     "pretest": "npm run-script lint",
     "test": "cross-env NODE_ENV=test nyc mocha-webpack --opts ./mocha-webpack.opts --webpack-config ./webpack.config.test.js -r test/setup.js test",
+    "coverage": "nyc report --reporter=text-lcov | coveralls",
```
jimporter: This is just reporting on the coverage we've already run in `npm test`, right? If so, I'd name this something like `coveralls`, since this script isn't actually doing the code coverage.

devinreams: Fair point 👍 I renamed it

PR #58, opened 2017-09-06
Improve our l10n code

jimporter:
This just moves stuff around and adds tests for our l10n code. Just look at the last 3 commits here.

Also: with this (and the dependent PRs) landed, we're at 94% test coverage. The only un-covered stuff are the entry points for our add-on, which probably aren't worth testing (at least not right now).

Closes #28

PR #59, opened 2017-09-07
The Great Renaming

jimporter:
This renames all our files to use hyphens instead of camelCase. That's more conventional and should avoid any problems on case-insensitive file systems. I also did the same for our CSS class names since that's the prevailing style in CSS.


devinreams:
After `npm install` and `npm run run` I got an error:

```
ERROR in ../node_modules/fluent-intl-polyfill/src/index.js
Module not found: Error: Can't resolve 'intl-pluralrules' in '/Users/dreams/src/lockbox-extension/node_modules/fluent-intl-polyfill/src'
 @ ../node_modules/fluent-intl-polyfill/src/index.js 9:0-58
 @ ./webextension/l10n.js
 @ ./webextension/manage/index.js
```

I removed `node_modules` and the package-lock 🔒then install'd and run'd again with success ✅ 

(any good way to prevent all that? is it expected I'd have to do that?)

jimporter:
That's weird. It shouldn't do that; `npm install` should be sufficient. (That change is technically from #58, not this PR though.)

jimporter:
@linuxwolf or @devinreams: review ping?

jimporter:
Merging without another review since all I did was rebase onto master.

PR #61, opened 2017-09-07
Add codecov to replace coveralls

devinreams:
I think this works both locally (if you have and set the token, which I did) and remotely (here on GitHub with the integration).

This PR will tell us if the latter does actually work...

https://codecov.io/gh/mozilla-lockbox/lockbox-extension


devinreams:
Cool, looks like Travis kicked off and sent the reports properly `after_success`:

https://travis-ci.org/mozilla-lockbox/lockbox-extension/builds/273100188?utm_source=github_status&utm_medium=notification

screen shot 2017-09-07 at 5 05 40 pm

devinreams:
>Do we actually want both codecov and coveralls? If the former works ok, I think we should probably just switch over to that.

Fine by me. My short experience with codecov has been much smoother than coveralls. I'm 👍 to fully switch.

I also _really_ don' enjoy like the comments (immediately disabled on coveralls). It's just a quick yml config change.

devinreams:
@jimporter Are you OK with this swap out?

I went ahead and did the same over on [lockbox-datastore](https://github.com/mozilla-lockbox/lockbox-datastore/pull/24)

package.json
```diff
@@ -31,7 +31,7 @@
     "run": "jpm run --addon-dir dist/ ${npm_config_jpm_runflags} --prefs jpm-prefs.json",
     "pretest": "npm run-script lint",
     "test": "cross-env NODE_ENV=test nyc mocha-webpack --opts ./mocha-webpack.opts --webpack-config ./webpack.config.test.js -r test/setup.js test",
-    "coveralls": "nyc report --reporter=text-lcov | coveralls",
+    "report-coverage": "npm run-script test && nyc report --reporter=lcov && codecov",
```
jimporter: I notice this changed to include `npm run-script test`. Is that necessary? Wouldn't that cause us to run the tests twice?

devinreams: Great catch. It should _not_ be necessary as test and coverage output will run by CI before this. Likely some residual from my local testing to make sure I actually had reports to work with. Thanks 🙇

devinreams:
OK two CI builds and report uploads seemed to work here: https://codecov.io/gh/mozilla-lockbox/lockbox-extension/commit/d0cf5cd76935a8bb4888ae7d47c6129d22c595ec/build

I'll finish painting this bikeshed for now...

Issue #26, opened 2017-07-25
Distribute Alpha extension on desktop


devinreams:
- [x] package script to build xpi
- [x] added to internal sign-xpi Jenkins pipeline (via Test Pilot and CloudOps)
- [x] confirmed signed properly
- [x] confirmed updates distributed
- [x] process documented
- [x] write 0.1.0-alpha changelog #76 #176 
- [x] publish 0.1.0-alpha tag and release on GitHub

devinreams:
Thinking about how to package, sign, and distribute this over at #62

sandysage:
Completed 10/16

PR #62, opened 2017-09-08
Add package signing and hosting instructions

### LINKED ISSUE #93: `npm run package` needs to create a file called addon.xpi
It looks like everything is happening well with the install and packaging scripts, except that you're naming the file `${npm_package_name}-${npm_package_version}.xpi`.  It should literally be `addon.xpi`

Thanks!

devinreams:
Starting to think about distribution stuff:

- [x] ability to package xpi (package)
- [x] ability to sign xpi
  - will be run through Mozilla Test Pilot deployment pipeline: https://github.com/mozilla/testpilot/blob/master/docs/development/hosting.md
  - [x] hook Jenkins up to this repo
- [x] set up update_url for self-hosted extension updates (will be provided by Test Pilot)
- [x] update release README instructions to match final of above ☝️ 

Addresses (but doesn't quite fully finish) #26 

Fixes #93


devinreams:
@linuxwolf @jimporter I'll need your help finishing this up. In theory I think everything is possible, but the big questions and todos are outlined in the description above ☝️

What are your initial thoughts?

devinreams:
Note to selves: there is user/secret keys in Mana and people we should probably talk to next.

I ran one set of keys through `jpm sign` and got 403: "You do not own this addon"

devinreams:
This morning I requested "internal signing" abilities: https://wiki.mozilla.org/Add-ons/InternalSigning

devinreams:
After chatting with @clouserw, I'm going to move things around a little in anticipation of matching this expected build process from Test Pilot: https://github.com/mozilla/testpilot/blob/master/docs/development/hosting.md

- add "production" branch for actual signed releases
- set `package` script to create addon.xpi (not lockbox-version.xpi)

devinreams:
@linuxwolf @jimporter This is ready for a quick review. Sorry for all the commit noise (stuff changed a lot along the way) please just check out the changed files. tldr:

1. make it so `package` creates addon.xpi
2. add instructions for packaging and releasing (via Test Pilot's Jenkins pipeline) on "production" branch
3. add an extension update URL (to Test Pilot) so updates are distributed

Read more about the deployment and hosting: https://github.com/mozilla/testpilot/blob/master/docs/development/hosting.md

If this looks OK to you, then I'll ask to get Jenkins hooked up here and test it out.

I already created "production" branch and set it to be properly "protected".

src/webextension/manifest.json.tpl
```diff
@@ -5,6 +5,13 @@
 
   "description": "{{description}}",
 
+  "applications": {
+    "gecko": {
+      "id": "lockbox@mozilla.com",
+      "update_url": "https://testpilot.firefox.com/files/lockbox@mozilla.com/updates.json"
```
jimporter: Is this the right place to put update URLs? I would have expected it to go in our `install.rdf`, since that's the "root" extension. I have no idea what's right here though, so if you've confirmed that this is ok with the Test Pilot folks, then this looks good to me.

devinreams: @jimporter Good catch and fair question. I poked around and saw other Test Pilot experiments with emebedded WebExtensions (including [containers](https://github.com/mozilla/testpilot-containers/blob/master/webextension/manifest.json#L15)) using the manifest.json for the updated URL. Either way, I've added a reminder to test and confirm this.

wresuolc: This is correct (assuming your ID is correct)

PR #65, opened 2017-09-11
Work with datastore asynchronously

linuxwolf:
This is an attempt to work with `lockbox-datastore` in an asynchronous manner.

This is probably not the right way to do it, but it minimized the changes needed.

### DIFF
--- a/src/webextension/background/datastore.js
+++ b/src/webextension/background/datastore.js
@@ -4,6 +4,11 @@
 
 import * as DataStore from "lockbox-datastore";
 
-const datastore = DataStore.create({});
+let datastore;
 
-export default datastore;
+export default async function openDataStore() {
+  if (!datastore) {
+    datastore = (typeof DataStore.open === "function") ? await DataStore.open() : DataStore.create();
+  }
+  return datastore;
+}
--- a/src/webextension/background/index.js
+++ b/src/webextension/background/index.js
@@ -2,20 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import datastore from "./datastore";
+import openDataStore from "./datastore";
 import initializeMessagePorts from "./message-ports";
 
 // XXX: For now, initialize the datastore on startup and then hook up the
 // button. Eventually, we'll have UX to create new datastores (and persist
-// existing ones).
-datastore.initialize().then(() => {
+// existing ones).\
+openDataStore().then(async(ds) => {
+  if (!ds.initialized) {
+    await ds.initialize();
+  }
   initializeMessagePorts();
+
   function openLockbox() {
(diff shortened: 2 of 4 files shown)


linuxwolf:
Latest push is simply rebasing from master.

src/webextension/background/datastore.js
```diff
@@ -4,6 +4,11 @@
 
 import * as DataStore from "lockbox-datastore";
 
-const datastore = DataStore.create({});
+let datastore;
 
-export default datastore;
+export default async function runit(fn) {
```
jimporter: Nit: I'd call this function something like `withDatastore`. I'm also not sure about passing a function to it; wouldn't it be better to simply return the datastore object? Then it's easier to do stuff like use `Promise.all()` with this function.

src/webextension/background/index.js
```diff
@@ -8,7 +8,10 @@ import initializeMessagePorts from "./message-ports";
 // XXX: For now, initialize the datastore on startup and then hook up the
 // button. Eventually, we'll have UX to create new datastores (and persist
 // existing ones).
-datastore.initialize().then(() => {
+datastore(async ds => {
```
jimporter: This is something we can discuss, but I generally prefer to put parens around the arg for single-arg arrow functions in JS; I think it's a more-consistent style, and probably slightly safer. `a => b` looks a lot like `a >= b`, even though they're *very* different.

I'm not married to that style, but I think it'd be good if we had a rule for it either way (which we could add to eslint).

src/webextension/background/datastore.js
linuxwolf: I changed it to `openDataStore()` and returning a Promise.

src/webextension/background/index.js
linuxwolf: I changed them to all use parens.

devinreams:
@linuxwolf Based on your latest commit, do you want a final review from @jimporter before merging?

linuxwolf:
@jimporter please re-review.  I hope I've addressed your comments.

src/webextension/background/datastore.js
```diff
@@ -4,6 +4,11 @@
 
 import * as DataStore from "lockbox-datastore";
 
-const datastore = DataStore.create({});
+let datastore;
 
-export default datastore;
+export default async function openDataStore() {
+  if (!datastore) {
+    datastore = (typeof DataStore.open === "function") ? await DataStore.open() : DataStore.create();
```
jimporter: We could probably wait to change this until we update our dependency on lockbox-datastore. We'll want to change this line anyway once we update, since we don't care about backwards-compatibility in this way. It doesn't matter too much for now though...

linuxwolf: 🐔 <==> 🥚

linuxwolf:
Merging without review -- simply rebased and squashed.

PR #66, opened 2017-09-11
disable codecov commenting 

devinreams:
in codecov docs, false means off: https://docs.codecov.io/v4.3.6/docs/pull-request-comments

screen shot 2017-09-11 at 1 57 19 pm

(waiting to make sure bot doesn't actually comment here)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,1 +1,0 @@
-comment: false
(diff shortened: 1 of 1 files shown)


devinreams:
Actually, let's remove `.codecov.yml` entirely since we can add it in the (newly discovered) Team YAML which will apply to all our repos:

See it: https://codecov.io/account/gh/mozilla-lockbox/yaml

```yaml
comment: false
```

Read more: https://docs.codecov.io/v4.3.6/docs/codecov-yaml

PR #68, opened 2017-09-11
(WIP) Set up Travis to push tags as GitHub releases with binaries

devinreams:
This is to follow along with prep going on at #62 and #26 for release distribution and packaging.

@jimporter as we quickly discussed, it'll be nice to automatically get Travis throwing the build/package output directly onto GitHub. Seems like there's an answer for that.

Here's a quick proof of concept:

- after CI success, it runs `package`
- the result of package (the XPI) is sent to GitHub as a release (hard-coded file path right now)
- the intention being to do this on tags, so we can bump versions and then have a resulting binary uploaded for _anyone_ to just download and run
  - it's scoped just to this branch for now so I'm not accidentally creating a ton of tags/releases

Some reading materials:

- GitHub releases: https://help.github.com/articles/creating-releases/
- Travis CI Releases integration: https://docs.travis-ci.com/user/deployment/releases/
- encrypted keys (no that's not my API key in plaintext): https://docs.travis-ci.com/user/encryption-keys/

Todo:

- [x] set up releases to go through Travis
  - [x] change branch back to default / master once ready
  - [x] change file name from hard-coded?
- [x] add first tag and release
  - reminder to self at #62 to change the Travis `after_success` to run packagesign (will fail now)

@linuxwolf @jimporter @m8ttyB any initial reactions, concerns, suggestions with this approach?

PR #69, opened 2017-09-12
Add template replace script to use package.json values in extension configs

devinreams:
@jimporter @linuxwolf here's a small script to find/replace tokens in templates (just for install.rdf and manifest.json for now).

Not sure if that's the right approach but wanted to get something rolling here so a release process (ie: one-step bump version in package.json) has less potential for mistakes or inconsistencies. 

What do you think?

Note: PR is based off my existing `staging` branch at #68 (related to #26)


devinreams:
I like the un-pythoning proposed at #72 😬 but reading @linuxwolf and @jimporter comments I realize we're already doing templates with webpack, so we should just do them the same way.

I may try (and fail) to hack at doing that approach a little later on then. But, if anyone wanted to try it first, feel free to say so and run with it.

jimporter:
I'll take on the templatizing stuff. It shouldn't take too long...

devinreams:
>I'll take on the templatizing stuff.

Thanks @jimporter. I'll close this PR then as it's served its purpose (discussion, decision).

PR #72, opened 2017-09-12
Remove python

hildjj:
Suggest template replacement in javascript instead of python?


devinreams:
I like it. 👍 I'm gonna merge it into my branch to keep the discussion and history...

..and then probably redo it all. 😔

PR #73, opened 2017-09-12
verify message-port sender is only the current extension

linuxwolf:
This code assumes all valid MessagePort messages only come from our extension, and nowhere else.  Exploring with `npm run run` seemed to prove that out ... so far.

### DIFF
--- a/src/webextension/background/message-ports.js
+++ b/src/webextension/background/message-ports.js
@@ -22,6 +22,11 @@ export default function initializeMessagePorts() {
   });
 
   browser.runtime.onMessage.addListener(async function(message, sender) {
+    // ensure only this extension can send itself messages
+    if (sender.id !== browser.runtime.id) {
+      throw new Error(`unauthorized sender "${sender.id}"`);
+    }
+
     switch (message.type) {
     case "list_items":
       return {items: Array.from((await datastore.list()).values(),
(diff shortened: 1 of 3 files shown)


linuxwolf:
Note that `./test/background/message-ports-tests.js` went through a lot of changes that ultimately is almost all rearranging deck chairs.

jimporter:
Is this necessary? My understanding was that `onMessage` was for listening to messages from our add-on, and [`onMessageExternal`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/onMessageExternal) was for listening to messages from *other* add-ons.

linuxwolf:
When I was first reading through these docs, it wasn't clear `onMessage` was restricted to just the extension's sandbox; it's not clear to me to what degree web accessible javascript can access.  I could be overly paranoid, and we're not bundling any web accessible content today, but given the sensitivity of the data I'm far more inclined to be too cautious.

jimporter:
Let's ask the add-ons team. My understanding is that what we have on master now is all we need, but it'd be good to get an answer from the experts about what we need to be worried about.

devinreams:
>My understanding is that what we have on master now is all we need, but it'd be good to get an answer from the experts about what we need to be worried about.

@linuxwolf as the author of the PR would you prefer to reach out (if you haven't already)? Let me know if you'd like me to, instead.

linuxwolf:
Asking about and looking at various extensions docs more, this PR is overkill.  Web accessible content runs in the same sandbox as the content that loads it.

PR #74, opened 2017-09-12
Templatize install.rdf and manifest.json

jimporter:


webpack.config.babel.js
```diff
@@ -2,17 +2,24 @@
...
 
-const entries = {
-  "webextension/background": "./webextension/background/index.js",
-  "webextension/manage/index": "./webextension/manage/index.js",
+import JSONWebpackPlugin from "./json-webpack-plugin";
+import thisPackage from "./package.json";
+
+const addonInfo = {
+  name: thisPackage.title,
+  id: "lockbox@mozilla.com",
```
linuxwolf: this can also be in the package.json, as `extensionId` since `id` seems too generic.

webpack.config.babel.js
```diff
@@ -2,17 +2,24 @@
...
 import MinifyPlugin from "babel-minify-webpack-plugin";
-import path from "path";
 
-const entries = {
-  "webextension/background": "./webextension/background/index.js",
-  "webextension/manage/index": "./webextension/manage/index.js",
+import JSONWebpackPlugin from "./json-webpack-plugin";
+import thisPackage from "./package.json";
+
+const addonInfo = {
```
linuxwolf: A comment here that this is for the templates would be helpful to future us

webpack.config.babel.js
jimporter: `id` might be ok actually, since it's what `jpm` uses when it generates `install.rdf` from `package.json`...

PR #82, opened 2017-09-15
Fix extension name and addon icon

### LINKED ISSUE #79: Lockbox extension name and icon disappeared
Related to #74 

screen shot 2017-09-15 at 1 16 23 pm

screen shot 2017-09-15 at 1 17 17 pm

devinreams:
So minor. Fixes #79

- [x] name just needed to refer to caps-case "Title" in `package.json`
- [x] system addon is simplest to just have an `icon.png` at the top-level and it'll read that

### DIFF
--- a/webpack.config.babel.js
+++ b/webpack.config.babel.js
@@ -94,6 +94,7 @@ export default {
       {from: "bootstrap.js"},
       {from: "webextension/**/*.ftl"},
       {from: "webextension/icons/*"},
+      {from: "icon.png"},
     ]),
     new webpack.DefinePlugin({
       "process.env": {
--- a/webpack.config.babel.js
+++ b/webpack.config.babel.js
@@ -94,8 +94,10 @@ export default {
       {from: "bootstrap.js"},
       {from: "webextension/**/*.ftl"},
       {from: "webextension/icons/*"},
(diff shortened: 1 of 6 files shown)


webpack.config.babel.js
```diff
@@ -94,6 +94,7 @@ export default {
       {from: "bootstrap.js"},
       {from: "webextension/**/*.ftl"},
       {from: "webextension/icons/*"},
+      {from: "icon.png"},
```
jimporter: Instead of symlinking `icon.png` to `webextension/icons/lock.png`, let's just copy the file there in webpack:

```js
{from: "webextension/icons/lock.png", to: "icon.png"}
```

devinreams: Yeah, good thinking. I had tried that, but the rule before it seemed to supersede the second rule. No matter what order I put two rules in, the copy couldn't run on the same file twice. Maybe I'm being dense and missed something, though.

devinreams:
I found the behavior to "only copy once" was by design (when file encountered it goes into a cache) and we can force the copy to happen again with `copyUnmodified: true`. @jimporter how's that looking now?

jimporter:
Hmm, let's just symlink it then.

Issue #80, opened 2017-09-15
Front-end search interface


devinreams:
Placeholder for adding text search to the user interface

- "live" filter updates list value immediately
- acceptable values: all characters

PR #83, opened 2017-09-19
Add a filter to search for items

jimporter:
This PR adds usernames to the summary list on the left and a filter that searches over the item's title and username (it splits the query on spaces and case-insensitively searches for each word individually for now).

This could still use some improvements, although it might be good enough to land already. Some stuff we should resolve (eventually) though:

* It can be confusing if the currently-selected item is filtered out; not sure what the best thing to do here is. Maybe deselect it?
* Maybe use something like Redux Forms?
* The styling of the search widget could use improvement.

lockbox-alpha

### DIFF
--- a/src/webextension/manage/components/item.js
+++ b/src/webextension/manage/components/item.js
@@ -2,20 +2,32 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { withLocalization } from "fluent-react";
 import PropTypes from "prop-types";
 import React from "react";
 
 import styles from "./item.css";
 
-export default function Item({name, selected, onClick}) {
+function Item({title, username, selected, getString, onClick}) {
   const attrs = (selected) ? {"data-selected": true} : {};
   return (
-    <li className={styles.item} {...attrs} onClick={onClick}>{name}</li>
+    <li className={styles.item} {...attrs} onClick={onClick}>
+      <div>
+        {title || getString("item-summary-no-title")}
+      </div>
+      <div className={styles.subtitle}>
+        {username || getString("item-summary-no-username")}
+      </div>
+    </li>
   );
 }
 
 Item.propTypes = {
-  name: PropTypes.string.isRequired,
-  selected: PropTypes.bool.isRequired,
+  title: PropTypes.string,
+  username: PropTypes.string,
+  selected: PropTypes.bool,
+  getString: PropTypes.func.isRequired,
   onClick: PropTypes.func.isRequired,
 };
+
+export default withLocalization(Item);
--- a/src/webextension/manage/containers/item-filter.js
+++ b/src/webextension/manage/containers/item-filter.js
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Localized } from "fluent-react";
+import PropTypes from "prop-types";
+import React from "react";
+import { connect } from "react-redux";
+
+import { filterItems } from "../actions";
+import { parseFilterString } from "../filter";
+import FilterInput from "../../widgets/filter-input";
+
+function ItemFilter({dispatch}) {
+  return (
+    <Localized id="item-filter">
+      <FilterInput placeholder="fILTEr…" onChange={(value) => {
+        dispatch(filterItems(parseFilterString(value)));
+      }}/>
+    </Localized>
+  );
+}
+
+ItemFilter.propTypes = {
+  dispatch: PropTypes.func.isRequired,
+};
+
+export default connect()(ItemFilter);
--- a/src/webextension/widgets/filter-input.js
+++ b/src/webextension/widgets/filter-input.js
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import PropTypes from "prop-types";
+import React from "react";
+
+import styles from "./filter-input.css";
+
+export default class FilterInput extends React.Component {
+  static get propTypes() {
+    return {
+      onChange: PropTypes.func,
+      value: PropTypes.string,
+    };
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      value: props.value || "",
+    };
+  }
+
+  updateValue(value) {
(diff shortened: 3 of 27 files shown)


devinreams:
@jimporter Looks good to me. I just did some simple functional tests:

- added "aaa" no username
- added "bbb" no username
- added "bbb" with "devin" username
- searched for z (no results) ✅ 
- searched for a (1 result) ✅ 
- searched for b (2 results) ✅ 

@m8ttyB is this the stage in development of a feature you'd like to join in and help, capture stuff for a test plan, check accessibility, etc.?

m8ttyB:
@devinreams yes, this is good stage. Is there a spec that we're targeting with regard to functionality? 

@jimporter ++ on the included tests.

devinreams:
>Is there a spec that we're targeting with regard to functionality? Specifics including what are acceptable and unacceptable values.

@m8ttyB Not to much detail, no. This corresponds to the very light detail at #80 to simply add a front-end text search interface for the list items.

My knee-jerk reaction is acceptable values are any characters. Can you please work with @jimporter to do a first pass definition of this (and anything else you think we should capture/test)?

devinreams:
> The input field helper text is filter, I'd prefer search.

@m8ttyB Agree on "Search" verbiage (which satisfies the user need by filtering the list).

>can you add an aria role="search" to the input search/filter form?

`/me reads up on ARIA`

Good call. I'll add a review "change requested" in the spot to make it easy for @jimporter to spot (among any others from you or @linuxwolf).

src/webextension/manage/containers/item-filter.js
```diff
@@ -0,0 +1,28 @@
...
+import { connect } from "react-redux";
+
+import { filterItems } from "../actions";
+import { parseFilterString } from "../filter";
+import FilterInput from "../../widgets/filter-input";
+
+function ItemFilter({dispatch}) {
+  return (
+    <Localized id="filter-items">
+      <FilterInput placeholder="fILTEr…" onChange={(value) => {
```
devinreams: As requested by @m8ttyB, please add the ARIA role="search".

src/webextension/manage/locales/en-US.ftl
```diff
@@ -1,4 +1,9 @@
 add-item = Add Item
```
devinreams: Sorry, I merged #87 so this file needs to get moved now.

src/webextension/widgets/filter-input.js
```diff
@@ -0,0 +1,49 @@
...
+    // eslint-disable-next-line no-unused-vars
+    const {onChange, value, ...props} = this.props;
+    return (
+      <div className={styles.filterInput}>
+        <span className="browser-style">
+          <input type="text" {...props} value={this.state.value}
+                 onChange={(e) => this.updateValue(e.target.value)}/>
+        </span>
+        <button type="button" className="browser-style"
+                onClick={() => this.updateValue("")}>X</button>
```
devinreams: Can you consider "Clear" or other meaningful language here for now? We can always swap the the text for an icon later.

devinreams: Is it possible to hide this button until there's a non-placeholder value in the input?
(i.e.: interface doesn't suggest you can "Clear" until there's something to clear...)

src/webextension/manage/components/item.js
```diff
@@ -2,20 +2,32 @@
...
+function Item({title, username, selected, getString, onClick}) {
   const attrs = (selected) ? {"data-selected": true} : {};
   return (
-    <li className={styles.item} {...attrs} onClick={onClick}>{name}</li>
+    <li className={styles.item} {...attrs} onClick={onClick}>
+      <div>
+        {title || getString("item-summary-no-title")}
+      </div>
+      <div className={styles.subtitle}>
+        {username || getString("item-summary-no-username")}
```
devinreams: This is a nice touch 👍

src/webextension/manage/containers/item-filter.js
devinreams: Also a minor nit worth considering: a placeholder of "Search..." may be more meaningful to the user (Search = what do they want to do, Filter = how that happens). Subject to change either way, I suppose...

src/webextension/widgets/filter-input.js
```diff
@@ -0,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import PropTypes from "prop-types";
+import React from "react";
+
+import styles from "./filter-input.css";
+
+export default class FilterInput extends React.Component {
```
devinreams: While its fresh, is now an good time to consider adding jsdoc comments? Or do you want to go back through everything separately and catch it all at once?

devinreams:
@jimporter Whoops, sorry if I was premature on my review comments ☝️ I just realized you didn't explicitly request any reviewers (yet).

devinreams:
>It can be confusing if the currently-selected item is filtered out; not sure what the best thing to do here is. Maybe deselect it?

I tried a few other apps/sites that do filtering like this, the behavior tends to be: de-select whatever you were looking at and select the first item in the filtered results list.

So if I have:

- A
- B *selected*
- C

And type "C" into the filter, I just get:

- C *selected*

But, that also checks and prompts the user if there were unsaved changes on "B".

src/webextension/widgets/filter-input.js
jimporter: It's possible, but I'm just going for basic right now. I expect this widget will be completely redesigned to have better styling later.

src/webextension/manage/containers/item-filter.js
jimporter: "Filter" is the word that makes more sense to me... when I see "Search", I usually think "this query will run when I hit enter", and when I see "Filter", I think "this query will run as I type".

src/webextension/widgets/filter-input.js
devinreams: >It's possible, but I'm just going for basic right now

Fair enough 👌

src/webextension/manage/containers/item-filter.js
devinreams: >when I see "Search", I usually think "this query will run when I hit enter"

That's likely an advanced / technical perspective you're looking at that from but I see your point. 😉 Like I said, minor nit and likely to change again so feel free to ignore me entirely...

jimporter: If it helps, here's some UX discussion about it: https://ux.stackexchange.com/questions/4752/search-vs-filter-what-is-the-difference

src/webextension/widgets/filter-input.js
jimporter: Per our Vidyo chat, let's do JSDocs after alpha.

PR #85, opened 2017-09-19
Update eslint

jimporter:
Now that <https://bugzilla.mozilla.org/show_bug.cgi?id=1395879> is fixed, we can update to the latest version of eslint (and plugins). I also updated the other out-of-date deps while I was here.


jimporter:
Note: if you pull this to try it, you may need to remove `node_modules` and reinstall. Webpack got fussy for me...

PR #87, opened 2017-09-19
Rearrange locales to match expected directory structure for Pontoon

devinreams:
@jimporter as mentioned, we'll want to hook up Pontoon to localize the extension strings and it expects every locale to have its own sub-directory:

https://developer.mozilla.org/en-US/docs/Mozilla/Implementing_Pontoon_in_a_Mozilla_website

>Individual locale folders must be located at the same nesting level of the directory tree. Source locale needs to be called templates, en or en-US. You may want to create a dedicated locales folder for this purpose. Locale code must not be part of the file name.

I went ahead and testing moving `locales/en-US.ftl` to `locales/en-US/extenstion.ftl` and this seems to work OK. Do you have any issues with this change or approach?


jimporter:
We should probably be putting this in `src/` or `src/webextension/`, since there will eventually be localization in areas other than the management UI.

devinreams:
@jimporter Good point. I moved them up into `src/webextension`.

How's that look now?

src/webextension/manage/index.js
```diff
@@ -20,7 +20,7 @@ initializeMessagePorts(store);
 
 ReactDOM.render(
   <Provider store={store}>
-    <AppLocalizationProvider baseDir="locales" availableLocales={["en-US"]}
+    <AppLocalizationProvider baseDir="/locales" availableLocales={["en-US"]}
```
jimporter: With `baseDir` defaulting to `"/locales"` in the class definition above, we can remove that prop from here. Then we should also add `bundle="manage"` so our provider knows what `.ftl` file to load.

devinreams:
>Looks good overall; just a few small changes.

@jimporter Good suggestions, thanks.

I _think_ I handled everything as expected but please check and give me feedback. 😏 

>Once this lands, I think I'll try to add something to eliminate the need to manually specify the list of available languages. Hopefully I can also add some error handling so we don't completely barf on missing locale files.

Great ideas 👍

src/webextension/l10n.js
```diff
@@ -9,16 +9,16 @@ import { LocalizationProvider } from "fluent-react";
...
   return { [locale]: messages };
 }
 
-async function createMessagesGenerator(baseDir, currentLocales) {
+async function createMessagesGenerator(baseDir, currentLocales, stringBundle) {
   const fetched = await Promise.all(
-    currentLocales.map((x) => fetchMessages(baseDir, x))
+    currentLocales.map((x) => fetchMessages(baseDir, x, stringBundle))
   );
   const bundle = fetched.reduce(
```
jimporter: Nit: let's call this variable `bundles` now, since it's technically a collection of bundles (one for each of `currentLocales`).

devinreams: Good idea.

That also lets me change `stringBundle` to `bundle`, for consistency and clarity. 👍

PR #88, opened 2017-09-20
Add documentation.js and configure mkdocs

devinreams:
In the same vein as https://github.com/mozilla-lockbox/lockbox-datastore/pull/32 and discussed here https://github.com/mozilla-lockbox/lockbox-datastore/issues/18, this PR:

- [x] uses `documentation.js` to generate docs from the webextension code
- [x] outputs it as markdown in `docs`
- [x] still allows other documentation / markdown files to exist in the directory for reference (`index.md` is a simplistic start)
- [x] configures `docs` in a way it can also be source for mkdocs to build our GitHub Pages site

I like this approach because it makes `/docs` the one-stop-shop for all docs, documentation.js can make legible markdown for us, and then mkdocs turns around and translates all that into a website.

Still needs:

- [ ] real jsdoc code comments
- [x] hook up to Travis to run the `doc` script and deploy to gh-pages

What do you think about this approach so far @jimporter? Any strong opinions or hesitation?


jimporter:
Is there a reason we're using Jekyll instead of MkDocs? I tend to prefer MkDocs because it's a bit simpler and has really great Markdown support thanks to all the plugins that the Python implementation has.

devinreams:
>Is there a reason we're using Jekyll instead of MkDocs? 

@jimporter initial idea was it has built-in support on GitHub (Pages) so we don't have to check in HTML and other assets: 
https://help.github.com/articles/using-jekyll-as-a-static-site-generator-with-github-pages/

They're mostly a wash but the idea here is actually "simpler": just put markdown in docs, add front matter to markdown, Jekyll / GitHub pages auto builds and handles the rest all on master, no extra scripts or build processes, no extra branches or commits.

jimporter:
The hooks from our CI are simpler for Jekyll, but what I was primarily concerned about was configuration and such. MkDocs does a lot for us, and is more oriented around building documentation sites specifically.

One thing I'd like to do eventually is host multiple versions of our docs, one for each release. I *think* that'll be easier with MkDocs than Jekyll. (MkDocs doesn't currently support this, but I'm an occasional contributor to the project and I have some ideas for how we'd do this in a good way.)

If we stick with MkDocs, we can also deploy pretty easily to ReadTheDocs if we decide to go that route; MkDocs is one of their supported documentation types. ReadTheDocs looks like it's pretty popular for other Mozilla projects.

I suppose it doesn't matter too much at this point, but I do think that MkDocs will be the better choice in the long term.

devinreams:
>I suppose it doesn't matter too much at this point, but I do think that MkDocs will be the better choice in the long term.

Nice, thanks for sharing. Sticking with mkdocs is no problem by me.

Assuming everything else in this PR is OK by you, we're just one revert away from going back through MkDocs.

@jimporter Want me to do that and then hook up Travis to run the doc+commit scripts?

devinreams:
OK I went back to mkdocs and moved our existing docs in here so everything is in one place.

`mkdocs build && mkdocs serve`

screen shot 2017-09-21 at 12 46 42 pm

@jimporter are you OK with the direction this is headed? If so I'll go hook up the deploy scripts so changes to code on master will update the api docs _in addition to_ pushing the mkdocs site to `gh-pages`.

README.md
```diff
@@ -7,7 +7,7 @@ Lockbox for Firefox is a work-in-progress extension for Firefox to improve upon
 Firefox's built-in password management. If you're interested, you should
 probably come back later when we have more to show!
 
-## Quick Start
+## [Quick Installation](install-link)
```
jimporter: If you're using a footnoted link like this, you need to use square brackets instead of parens.

docs/api.md
```diff
@@ -0,0 +1,65 @@
+<!-- Generated by documentation.js. Update this documentation by updating the source code. -->
```
jimporter: Should we add this to .gitignore? It's a generated file.

package.json
```diff
@@ -25,6 +25,7 @@
     "addon"
   ],
   "scripts": {
+    "doc": "documentation build src/webextension/** -f md -o docs/api.md",
```
jimporter: Should we have a script for actually running MkDocs here? (Bonus if we run it and deploy our docs to Github in Travis CI.)

README.md
devinreams: >you need to use square brackets

Oh shoot, good catch. Thank you...

docs/api.md
devinreams: I personally liked having the resulting markdown available if someone, didn't want/couldn't use mkdocs to generate it themselves, or couldn't get to the web to see the resulting output.

I don't have a strong opinion either way but having it checked in makes things easier to find and in the HTML generation pipeline...

package.json
devinreams: Yup, sorry I didn't get that far. I wanted to make sure we wer eon the right track. I've got this further along over in lockbox-datastore.

I will finish up over there (assuming @linuxwolf is cool with the approach too) and then backport the final bits over here...

docs/api.md
jimporter: If we have this hooked up to our CI, people can just check out our Github page, which should be sufficient for 99% of people. (For the 1%, they don't actually need MkDocs, since they can just run `npm run doc` and get the Markdown file to look at.)

Mainly, I just don't want this file to get out of sync (and I'm not a huge fan of automatic post-commit changes or things of that nature).

jimporter: Also, at some point I'd like to create a fix for mkdocs/mkdocs#193, which will let us host multiple versions of our docs in one place. I have some ideas for that, but it's not a super-high priority for me at the moment.

devinreams:
@jimporter OK this has been wrapped up 🎁 over at lockbox-datastore and appears to work well. ✅ 

>Should we have a script for actually running MkDocs here? (Bonus if we run it and deploy our docs to Github in Travis CI.)

Yup, this is now done at CI "deploy" step (only if on master branch).

screen shot 2017-09-25 at 4 03 59 pm

Please take one more look and see if you notice anything we should address before merging?

devinreams:
@jimporter Do you have a moment to check out this approach before next week?

I think we have it in a good spot. You can see it in action at: https://mozilla-lockbox.github.io/lockbox-datastore/

mkdocs.yml
```diff
@@ -1,2 +1,14 @@
...
+repo_url: https://github.com/mozilla-lockbox/lockbox-extension
+
+pages:
+- 'Introduction': 'index.md'
+- 'Installing': 'install.md'
+- 'User Guide': 'user-guide.md'
+- 'Contributing': 'contributing.md'
+- 'Code of Conduct': 'code_of_conduct.md'
+- 'API Guide': 'api.md'
+- 'Releases': 'releases.md'
```
jimporter: We could probably do this later, but it might make sense to nest some of these pages, e.g. under a "Developer" section.

mkdocs.yml
```diff
@@ -1,2 +1,14 @@
-site_name: Lockbox Docs
+site_name: Lockbox Extension
+site_description: Documentation for the Firefox extension
 theme: readthedocs
+repo_name: GitHub
```
jimporter: I don't think we need this, since it should auto-detect from the `repo_url`.

mkdocs.yml
devinreams: I thought about that but also liked keeping the two levels of nav and sub nav.

Either way, "docs polish" is on my list to discuss more next week. 👍

PR #90, opened 2017-09-20
Tweak linting config to use .eslintignore and lint bonus files

pdehaan:


json-webpack-plugin.js
```diff
@@ -11,7 +11,7 @@ export default class JSONWebpackPlugin {
       template: undefined,
       filename: undefined,
       data: {},
-      ...options
+      ...options,
```
pdehaan: This file wasn't being linted previously. ¯\\\_(ツ)\_/¯

PR #92, opened 2017-09-20
Add more fields to the management UI

### LINKED ISSUE #81: Update login CRUD interface: title, other data
Right now site is the title, we should explicitly capture and expose origin separately to match the datastore expectations

jimporter:
This resolves #81 and is based on #83 (so ignore the first two commits).

### DIFF
--- a/src/webextension/manage/components/item-details.css
+++ b/src/webextension/manage/components/item-details.css
@@ -3,4 +3,20 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 .item-details {
+  display: grid;
+  grid-template-columns: max-content 200px;
+  grid-column-gap: 1em;
+}
+
+.item-details > label {
+  display: contents;
+}
+
+.item-details > label > span:first-child {
+  /* This lines up our labels with the text in <input> elements. */
+  margin-top: 3.5px;
+}
+
+.item-details > p > button + button {
+  margin-inline-start: 1ch;
 }
--- a/src/webextension/manage/components/item-details.css
+++ b/src/webextension/manage/components/item-details.css
@@ -20,3 +20,7 @@
 .item-details > p > button + button {
   margin-inline-start: 1ch;
 }
+
+.item-details textarea {
+  resize: none;
+}
--- a/src/webextension/manage/containers/current-item.js
+++ b/src/webextension/manage/containers/current-item.js
@@ -18,10 +18,12 @@ function unflattenItem(item, id) {
   return {
     id,
     title: item.title,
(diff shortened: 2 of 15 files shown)


jimporter:
Here's what it looks like (finally our UI is less grody!):

Screenshot

test/manage/components/item-details-test.js
```diff
@@ -21,6 +21,30 @@ function simulateTyping(wrapper, value) {
 }
 
 describe("<ItemDetails/>", () => {
+  const blankFields = {
```
devinreams: Nice, this is clearer to me now :+1:

src/webextension/manage/components/item-details.css
```diff
@@ -3,4 +3,20 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 .item-details {
+  display: grid;
```
devinreams: Cool! Thanks for using CSS grid 🥇

src/webextension/manage/containers/current-item.js
```diff
@@ -18,28 +18,35 @@ function unflattenItem(item, id) {
...
       password: item.password,
+      notes: item.notes,
     },
   };
 }
 
 function flattenItem(item) {
   return {
     title: item.title,
+    origin: item.origins[0] || "",
```
devinreams: Nice, so if I understand correctly: we're just capturing one string for now and putting it into the array to match the expected datastore model.

linuxwolf: yes, exactly.  We can expand this later.

src/webextension/widgets/text-area.js
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
linuxwolf: I recommend there be a .css for textrea that specifies `resize: none;`.



I really don't think we'll have a valid scenario where a resizable textarea (without resizing its container) is desired.

src/webextension/manage/filter.js
```diff
@@ -0,0 +1,23 @@
...
+    return [];
+  }
+  return filter.split(/\s+/).map((i) => i.toLocaleLowerCase());
+}
+
+export function filterItem(filter, item) {
+  for (let i of filter) {
+    if (!item.title.toLocaleLowerCase().includes(i) &&
+        !item.username.toLocaleLowerCase().includes(i)) {
+      return false;
```
linuxwolf: It would be really helpful if you could filter/search on origin, too.

We can go back-and-forth on whether its visible or not in the summary ...

Issue #53, opened 2017-09-05
Extension item details: click to reveal and copy (login, password)


devinreams:
A quick search turned up some straight forward npm packages for both but not sure if they're worth incorporating as a dependency vs. rolling our own.

- click to reveal (input type password vs text)
  - https://www.npmjs.com/package/react-password-mask

- click to copy (grab input contents to clipboard)
  - https://www.npmjs.com/package/react-copy-to-clipboard

Any strong opinions one way or another (build ourselves vs package) @jimporter @linuxwolf?

jimporter:
Both seem simple enough that it's probably less effort for us to roll our own; that way, it'll be easier for us to make tweaks as needed.

devinreams:
@jimporter Agreed. Wanna run at it after lock/unlock?

linuxwolf:
This should include the ability to copy the username, too.

PR #98, opened 2017-09-21
Create a scrollable, keyboard-accessible list

jimporter:
(Note: ignore the first 4 commits since they're in other PRs.)

This adds a generic `<ScrollingList/>` widget that we can use for our left-hand pane. It also adds keyboard accessibility to the list so users *should* be able to navigate the entire UI without touching their mouse.

### DIFF
--- a/src/webextension/manage/components/app.js
+++ b/src/webextension/manage/components/app.js
@@ -17,7 +17,9 @@ export default function App() {
       <aside>
         <ItemFilter/>
         <AllItems/>
-        <AddItem/>
+        <div>
+          <AddItem/>
+        </div>
       </aside>
       <article>
         <CurrentItem/>
--- a/src/webextension/manage/components/item-details.js
+++ b/src/webextension/manage/components/item-details.js
@@ -13,7 +13,7 @@ import TextArea from "../../widgets/text-area";
 import styles from "./item-details.css";
 
 // Note: ItemDetails doesn't directly interact with items from the Lockbox
-// datastore. For that, please consult <../containers/currentItem.js>.
+// datastore. For that, please consult <../containers/current-item.js>.
 
 export default class ItemDetails extends React.Component {
   static get propTypes() {
--- a/src/webextension/manage/components/app.css
+++ b/src/webextension/manage/components/app.css
@@ -9,22 +9,18 @@ body {
 .app {
   display: grid;
   grid-template-columns: 1fr 5fr;
-  grid-template-rows: auto;
+  grid-template-rows: 100vh;
   grid-gap: 1em;
-  width: 100vw;
-  height: 100vh;
 }
 
 .app > aside {
   grid-column: 1;
   background-color: #f9f9fa;
-  padding: 0 
+  display: grid;
+  grid-template-rows: auto 1fr auto;
 }
 
 .app > article {
   grid-column: 2;
-}
-
-.app > article {
   margin: 1em;
 }
--- a/src/webextension/manage/components/app.css
+++ b/src/webextension/manage/components/app.css
@@ -15,6 +15,7 @@ body {
 
 .app > aside {
   grid-column: 1;
+  min-width: 0;
   background-color: #f9f9fa;
   display: grid;
   grid-template-rows: auto 1fr auto;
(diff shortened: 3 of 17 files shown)


jimporter:
@m8ttyB You were talking about a11y earlier. With this change (and the dependents PRs), the management UI should be fully keyboard-accessible (although I make no claims that the accessibility is *good* yet).

jimporter:
@m8ttyB You were talking about a11y earlier. With this change (and the dependents PRs), the management UI should be fully keyboard-accessible (although I make no claims that the accessibility is *good* yet).

jimporter:
Just realized one thing I forgot: the `<ScrollingList/>` should actually scroll when you use the arrows to select an item out of the current visible region.

jimporter:
Ok, list items scroll into view as needed. I also changed selecting items to work on mousedown to reduce a weird flicker when clicking on an unfocused `<ScrollingList/>`; I don't know if this is quite the right thing to do, so I'm willing to drop that commit entirely. Just let me know what you think...

src/webextension/manage/components/app.js
```diff
@@ -17,7 +17,9 @@ export default function App() {
       <aside>
         <ItemFilter/>
         <AllItems/>
-        <AddItem/>
+        <div>
```
devinreams: Good move 👍

jimporter:
@m8ttyB I'll do the form focusing in a separate PR. This one's already fairly complex.

m8ttyB:
@jimporter Sounds great to me 💯

pdehaan:
Also, unrelated to this PR, but should we make at least _one_ field in the Add form required? Currently I can add infinite number of new records where I don't fill in any content and all the results just say "(no title)" / "(no username)"

jimporter:
Eventually, passwords will get auto-generated (at least by default), so every item will have one unique value (not that we'd show the password in the item list). I think it's fine to let people have items with no title/username if they want, since it doesn't break anything.

src/webextension/manage/components/item-details.js
```diff
@@ -62,27 +61,42 @@ export default class ItemDetails extends React.Component {
...
-        </p>
-        <p>
-          <Localized id="item-password">
+        </label>
+        <label>
+          <Localized id="item-details-password">
             <span>pASSWORd</span>
           </Localized>
-          <Input type="text" value={this.state.password}
+          <Input type="text" name="password" value={this.state.password}
```
pdehaan: Should this be `input="text"` or `input="password"`. I was a bit surprised the first time I tried adding a new entry and my password was displayed in plaintext.
Not sure if we want to consider something like [**react-input-password**](https://www.npmjs.com/package/react-input-password) or [**react-password-mask**](https://www.npmjs.com/package/react-password-mask).

src/webextension/manage/components/item-details.js
```diff
@@ -62,27 +61,42 @@ export default class ItemDetails extends React.Component {
...
-        </p>
+        </label>
+        <label>
+          <Localized id="item-details-notes">
+            <span>nOTEs</span>
+          </Localized>
+          <TextArea name="notes" value={this.state.notes}
+                    onChange={(e) => this.setState({notes: e.target.value})}/>
+        </label>
+        <p></p>
```
pdehaan: Nit: An empty `<p></p>` feels weird. Not sure if we should wrap in a `<div>` instead to add some extra padding if we need it.

src/webextension/manage/components/item-details.js
```diff
@@ -62,27 +61,42 @@ export default class ItemDetails extends React.Component {
         e.preventDefault();
         onSave(this.state);
       }}>
-        <p>
-          <Localized id="item-site">
-            <span>sITe</span>
+        <label>
+          <Localized id="item-details-title">
+            <span>tITLe</span>
```
pdehaan: Personal curiosity, but why a nested `<span>` tag instead of using `<label htmlFor="xyz">tITLe</label>` and then giving an `id` attribute to the `<Input>` tags?

jimporter: IDs are global across the entire page, which is a leaky abstraction; if I started using IDs, I'd have to manage ID uniqueness across supposedly-independent components. We already have to do that with localization keys, but I hope to eliminate those issues at some point too. (We could use some sort of unique ID generator, but that's extra complexity for something I can get for free by simply avoiding IDs in the first place).

That said, I might be able to just give `<Localized>` some text.

src/webextension/manage/components/item-details.js
jimporter: It should be `input="text"`; hiding the password is #53.

jimporter:
I force-landed this since all I did was rebase to make sure CI passed.

PR #106, opened 2017-09-25
Add a view mode for the item details

jimporter:
The plan here is to change the item details so that we're not *always* editing them. From the "view" mode, you can click Edit or Delete, and then in "edit" mode, you can edit the fields, Save, or Cancel.

This will also make auto-focusing the first form field in the editor easier, since it won't interrupt you when navigating through the list of items. Finally, I added a "new item" placeholder in the item list, since I think that should make it clearer what's going on for users.

### DIFF
--- a/src/webextension/manage/components/item-details.js
+++ b/src/webextension/manage/components/item-details.js
@@ -7,107 +7,59 @@ import PropTypes from "prop-types";
 import React from "react";
 
 import Button from "../../widgets/button";
-import Input from "../../widgets/input";
-import TextArea from "../../widgets/text-area";
 
 import styles from "./item-details.css";
 
 // Note: ItemDetails doesn't directly interact with items from the Lockbox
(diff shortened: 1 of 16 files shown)


src/webextension/manage/components/item-details.js
```diff
@@ -7,107 +7,59 @@ import PropTypes from "prop-types";
...
+        </Localized>
+        <span>{fields.origin}</span>
+      </div>
+      <div className={styles.field}>
+        <Localized id="item-details-username">
+          <span>uSERNAMe</span>
+        </Localized>
+        <span>{fields.username}</span>
+      </div>
+      <div className={styles.field}>
```
devinreams: To confirm, the plan is removing password from display here in this PR and adding it back (obfuscated) as part of #53, right? (knowing its also available during edit mode)

jimporter:
Force merging since all I did was rebase.

PR #108, opened 2017-09-26
A wild Lockie appeared!

jimporter:
This adds a "home" view that we could eventually use to provide statistics about the datastore, important news, etc. However, for the time being, it's Lockie's home. The code for Lockie's speech bubble is pretty brittle and probably falls over for a lot of locales, but by the time we actually support other locales, we'll probably have something fancier in place.


jimporter:
Force-landing since I only rebased from the latest master.

PR #111, opened 2017-09-27
Copy and show/hide password

### LINKED ISSUE #53: Extension item details: click to reveal and copy (login, password)

jimporter:
This resolves #53. It's based on my other PRs, so just look at the last two commits.


src/webextension/locales/en-US/manage.ftl
```diff
@@ -38,6 +38,7 @@ item-details-title = Title
 item-details-origin = Origin
 item-details-username = Username
```
linuxwolf: I'd really like the ability to copy the username, too.

devinreams: >I'd really like the ability to copy the username, too.

Great point 👍 I have some complex usernames and a shortcut here too would be nice.

jimporter:
The actual failures are because Enzyme doesn't have some bits that `react-copy-to-clipboard` expects, so I'll have to shim it for now. A lot of this will probably be easier with Karma, but I'm waiting until after alpha for that, since it'd probably take a day or two to get everything working.

I just forgot to look at tests before I pushed this. vOv

jimporter:
Actually, the failures were because the test code to figure out what button to click was very brittle. I fixed that, so it should be good for the foreseeable future now.

jimporter:
I also added the ability to copy the username.

jimporter:
Force-landing, since all I did was rebase.

Issue #70, opened 2017-09-12
Define and capture quantitative user analytics in extension for Alpha


devinreams:
https://trello.com/c/2EIG9hqU/5-metrics

- [ ] document and publish https://github.com/mozilla-lockbox/lockbox-extension/pull/118
- [ ] hook up to extension https://github.com/mozilla-lockbox/lockbox-extension/pull/114

devinreams:
@jimporter @linuxwolf The requirements have been drafted by @irrationalagent and are going through one more internal review.

Next up: expect a PR here tomorrow at `docs/metrics.md` describing what we want to collect, for your review, with hopes to get some initial events actually entered and captured by telemetry in the Alpha.

PR #114, opened 2017-09-28
Attempt to register and record Event Telemetry

devinreams:
@jimporter @linuxwolf I was looking over the new add-on "Event Telemetry" stuff and attempted to register and record an event to pave the way for #70:

- https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html#registerevents
- https://blog.mozilla.org/data/2017/09/18/recording-new-telemetry-from-add-ons/

With this code I don't get any errors (from what I can tell) but I also don't see any "Events" in `about:telemetry`. 

Can you tell if I've done something wrong?

### DIFF
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -4,8 +4,21 @@
 
 /* eslint-disable no-unused-vars */
 
+Components.utils.import("resource://gre/modules/Services.jsm");
+
 function startup({webExtension}, reason) {
+  Services.telemetry.registerEvents("lockbox", {
+    "startup": {
+      methods: ["startup"],
+      objects: ["addon", "webextension"],
+    },
+    "click": {
+      methods: ["click"],
+      objects: ["browser_action"],
+    },
+  });
   webExtension.startup().then(() => {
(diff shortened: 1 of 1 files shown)


jimporter:
This should work if we add the following to `jpm-prefs.json`:

```js
  "toolkit.telemetry.enabled": true,
  "toolkit.telemetry.server": "https://127.0.0.1/telemetry-dummy/",
```

Not sure what, if anything, we should do about unit testing this code...

src/bootstrap.js
```diff
@@ -4,8 +4,21 @@
 
 /* eslint-disable no-unused-vars */
 
+Components.utils.import("resource://gre/modules/Services.jsm");
+
 function startup({webExtension}, reason) {
+  Services.telemetry.registerEvents("lockbox", {
```
jimporter: I wonder if this should go in global scope. We might want to have events for enabling/disabling the add-on, and `startup` is only run when the add-on is enabled.

devinreams:
@jimporter would you prefer I just close this branch and let you run with a separate one?

jimporter:
Sure, that'll probably be easiest.

PR #115, opened 2017-09-28
Add nsp for dependency vuln checking

pdehaan:
Current output:

```sh
$ npm run nsp

> lockbox@0.1.0-pre nsp /Users/pdehaan/dev/github/mozilla/lockbox-extension
> nsp check

(+) No known vulnerabilities found
```


pdehaan:
> What happens if it finds a vulnerability? It's not clear to me where the results get sent. Do we just have to check the Travis logs?

Currently (as of [**nsp@2.8.x**](https://github.com/nodesecurity/nsp)), if it finds a vulnerable module, it will exit with a non-zero status code, and break the Travis-CI build. At which point we'd either have to:

(a) Update the questionable module, if updated version is available.
(b) Create an [.nsprc](https://github.com/nodesecurity/nsp#exceptions) file which excludes that one  advisory (and then remove that exception once the module is updated in the future).
(c) Ignore any build breaking errors using `nsp check`'s optional [`--warn-only`](https://github.com/nodesecurity/nsp/blob/11d5635385cdb453f1201b2dec0898f0780d4838/usage/check.txt#L25-L27) flag, which will _"report errors, but always quit with an exit code of 0"_.

.travis.yml
```diff
@@ -8,6 +8,7 @@ script:
 
 after_success:
   - npm run codecov
+  - npm run nsp
```
pdehaan: Counter proposal (if we want to trade 2-3 seconds per build for some smaller diff churn).
Instead of running `npm run nsp` here, we can use npm 5's `npx` magic to just call `npx nsp check` here, then not have to touch package.json or package-lock.json.

**Pros:**
- Moves check purely to Travis or Circle CI.

**Cons:**
- Makes it a bit more challenging to run `nsp` check locally (unless you install it globally or do similar `npx` trick).
- Installs the `nsp` module "on-demand" each time the CI runs, so in addition to the automatic `npm install`, the packages would need to be downloaded again for each build and not get cached in package-lock.json or shrinkwrap or yarn.lock files.

jimporter:
> Currently (as of nsp@2.8.x), if it finds a vulnerable module, it will exit with a non-zero status code, and break the Travis-CI build.

I don't think that will happen with the current patch, since `nsp` is run during `after_success`; per the Travis docs:

> The exit code of `after_success`, `after_failure`, `after_script` and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as a failure.

If it were me, I'd set up another build via the build matrix and run `nsp` there, but perhaps that should be part of a larger refactor of our Travis builds, since they assume we're doing everything linearly right now.

jimporter:
Thanks for the patch!

PR #116, opened 2017-09-28
Prototyping First-time User Experience (+ Unlock)

linuxwolf:
This is my first attempt.  It's ugly, and it's wrong.  BUT, still would like eyes on it.

@jimporter @devinreams @changecourse @m8ttyB

### DIFF
--- a/src/webextension/background/views.js
+++ b/src/webextension/background/views.js
@@ -1,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const views = new Map();
+export async function openView(name) {
+  console.log(`opening view ${name} ...`);
+  let id = views.get(name);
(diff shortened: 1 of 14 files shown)


linuxwolf:
I'm working through the npm issues now ... will update the PR when addressed.

src/webextension/background/views.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  if (info) {
+    let { windowId } = info;
+    await browser.windows.update(windowId, {
+      focused: true,
+    })
+    info = await browser.tabs.update(id, {
+      active: true,
+    });
```
jimporter: I still think we should let users open multiple management tabs, since it makes it a lot easier to copy data back and forth between multiple entries (e.g. copying some notes over to a new entry).

linuxwolf: I admit this was another experiment in a mix of experiement.

We can talk about single vs. multiple (especially interested in what @changecourse  and @sandysage might think), but I am worried about users getting confused ... which almost always leads to users feeling less trust.

PR #118, opened 2017-09-29
Metrics for event recording

### LINKED ISSUE #70: Define and capture quantitative user analytics in extension for Alpha

irrationalagent:
Closes #70 

This file is currently under review by @rweiss (data steward) but probably won't change that much.

This is a wishlist type overview of the metrics we'd like to collect. A review by engineering to see what is actually feasible would be the first place to start.

In terms of implementation: since we are deploying as an embedded web extension we should be able to access this public API for event recording (via the add-on part of the code which I believe is mostly empty at this point):

https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/events.html#public-js-api

See the those docs as well as metrics.md for details on how to use the api. I've included example javascript that could be modified to register and record the events. Obviously all the variable names there will need to change. 

To check whether events are being recorded, type about:telemetry into the firefox browser bar. The events should land in `processes.dynamic.events` . You may have to check the raw json for them. 

In the meantime I'll see what I can figure out on my own.

I'd like to help as much as possible with this but realistically I'll need a decent amount of engineering help placing the calls to the API in the right place and referencing the correct objects. 

The good news is that once we are correctly recording the events locally, they will be sent off with the main ping and I can do all the work of pulling them out of the moz data pipe.


docs/metrics.md
```diff
@@ -0,0 +1,440 @@
...
+
+The **methods** for these events are `render` and `click` for rendering and interaction events, respectively.
+
+
+The **extra** field contains the user's FxA anonymous/random user id. For `lockbox_toolbar_firstrun` and `welcome` views this will be `null`.
+
+
+To register the rendering events:
+```javascript
+Services.telemetry.registerEvents("setup.render", {
```
jimporter: Do we want the event category to have "lockbox" in it somewhere? That would probably avoid confusion.

docs/metrics.md
```diff
@@ -0,0 +1,440 @@
...
+                    "done",
+                    null,
+                    {"fxa_uid": uid} ) // user has signed in by now);
+```                    
+
+
+To record (for example) a click on one of the buttons:
+
+```javascript
+Services.telemetry.recordEvent("lockbox_toolbar_firstrun.interaction",
```
jimporter: This should be `"setup.interaction"`, right?

devinreams:
@irrationalagent are you ready for us all to re-review this? 

If so, feel free comment here and/or to select "Reviewers" in the top left to explicitly signal that.

irrationalagent:
yes, ready to be reviewed now. thanks!

docs/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+                    null,
+                    {"fxa_uid": null} ) // user hasn't signed in yet);
+```                    
+### Toolbar Button Interaction Event
+At this stage of development, the Lockbox toolbar icon is the main entrypoint for using the Lockbox UI. There is only one event of concern here, a click on the icon. This event has its own **category** of `toolbar.interaction`. The **method** is click, and the object is **toolbar_button**.
+
+The event can be registered as follows:
+
+```javascript
+Services.telemetry.registerEvents("toolbar.interaction", {
```
linuxwolf: Strictly speaking, this event can only be recorded if a popup is **not** registered.

That said, we could assume this event occurred if a popup view is rendered and explicitly record a "toolbar.click" event when a popup view is rendered (in addition to or instead of other events), or infer this event if one of the popup view rendering events is recorded.

docs/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+```javascript
+Services.telemetry.registerEvents("new_item.interaction", {
+  "item_editor.click": {
+    methods: ["click", "focus"],
+    objects: ["title_field",
+        "origin_field",
+        "username_field",
+        "password_field",
+        "notes_field",
+        "save_entry_button",
```
linuxwolf: Correct me if I'm mistaken, but focus events for buttons do not happen if the mouse (or touch gesture) is used to activate a button.

Assuming the above statement is true, Is it still worth tracking "focus" events for buttons?

docs/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+                    "save_entry_button",
+                    null,
+                    {"fxa_uid": uid});
+```
+
+### Events Related to Submitting An Item Change from the Item Editors
+Events of **category** `item_change_submitted` record the actual submission of new item information to the datastore (either completely new entries or updates to existing items).
+
+
+In theory, these should be tied to equivalent events in the datastore, so that (for example) an `adding` event which is logged when a user submits a new item has a corresponding `added` event that gets logged when the item is successfully added to the datastore. This may need to change, but for simplicity's sake I've made it as close to the datastore events as possible (see next section). These events will be recorded as `item_change_submitted`.
```
linuxwolf: Consideration for the future:

* When autofill is implemented, we could still record an "item_change_submitted" event, adding a field to help disambiguate from the same activity in the editor; or a new event could be registered/recorded.

docs/metrics.md
irrationalagent: Since we already have a render event for the item list, I'll remove this event from the doc

docs/metrics.md
irrationalagent: I removed the focus events from the doc

irrationalagent:
I updated this based on today's meeting and comments here. See the TLDR section at the top of the doc for alpha priorities

docs/metrics.md
```diff
@@ -0,0 +1,435 @@
...
+```javascript
+Services.telemetry.registerEvents("event_category", {
+    "event_name": {
+        methods: ["click", ... ], // types of events that can occur
+        objects: ["a_button", ... ], // objects event can occur on
+        extra: {"key": "value", ... } // key-value pairs (strings)
+    }
+```
+
+For our purposes, we will use the `extra` field for a few purposes:
```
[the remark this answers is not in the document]
linuxwolf: (As further discussed IRL, a unique flow or action id does not appear to be necessary at this time.)

docs/metrics.md
```diff
@@ -98,11 +112,11 @@ For our purposes, we will use the `extra` field for a few purposes:
 
 -   To log the FxA user id of the client logging the event (e.g. `"fxa_uid": uid`)
 -   To log the UUID of the item that has been added or changed (e.g. `"item_id": UUID`)
--   To log the fields that are modified when an item is updated in the datastore (e.g. `"fields":"password,notes"`  (because the value has to be a string we will have to concat the fields that were updated somehow)
+-   To log the fields that are modified when an item is updated in the datastore (e.g. `"fields": "password,notes"`  (because the value has to be a string we will have to concat the fields that were updated somehow)
```
linuxwolf: Looking at this again, I note the backed has some quirks with what the user sees versus what's recorded.  The JSON structure (before encryption) looks something like (`...` added for data that's likely not relevant):

```javascript
{
  "id": <uuid>,
  "title": <user string>,
  "origins": [0..n<URL string>],
  "tags": [0..n<user token>],
  ...,
  "entry": {
    "kind": "login",
    "username": <user string>,
    "password": <user string>,
    "notes": <user string>
  }
  "
}
```

It would be convenient if we could allow for these quirks in the metrics model.  For example, instead of:

```
"fields": "title,origin,password,notes"
```

if we could instead emit:

```
"fields": "title,origins,login.pasword,login.notes"
```

linuxwolf:
So, if `"fxa_ui"` was not acceptable to the Telemetry API, would `"item_id"` be?  I think there's one or two `extra` members with underscore names, too.

Issue #25, opened 2017-07-21
npm run-script build fails on Windows expecting zip command


m8ttyB:
On Windows 10, `npm run-script build` throws an error due to `zip` not accessible from the terminal.

Note, `npm run-script run` successfully works.
```
21 error Windows_NT 10.0.15063
22 error argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run-script" "package"
23 error node v6.11.1
24 error npm  v3.10.10
25 error code ELIFECYCLE
26 error lockbox@0.1.0-pre package: `cd dist && zip -r ../lockbox-0.1pre.xpi *`
26 error Exit status 1
27 error Failed at the lockbox@0.1.0-pre package script 'cd dist && zip -r ../lockbox-0.1pre.xpi *'.
27 error Make sure you have the latest version of node.js and npm installed.
27 error If you do, this is most likely a problem with the lockbox package,
27 error not with npm itself.
27 error Tell the author that this fails on your system:
27 error     cd dist && zip -r ../lockbox-0.1pre.xpi *
27 error You can get information on how to open an issue for this project with:
27 error     npm bugs lockbox
27 error Or if that isn't available, you can get their info via:
27 error     npm owner ls lockbox
27 error There is likely additional logging output above.
28 verbose exit [ 1, true ]
```

pdehaan:
Wouldn't this be solved by switching from **jpm** to **web-ext** and then using [`web-ext build`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_build) to build the add-on?

jimporter:
@pdehaan My understanding is that `web-ext` doesn't work for classic add-ons with embedded WebExtensions, only for pure WebExtensions. (We don't currently do much of anything in the XPCOM side of our add-on, but we will when we add support for form filling.)

PR #121, opened 2017-10-02
Output addon.env at build for Jenkins pipeline

devinreams:
This way pipeline scripts looking for a package.json (addon) _or_ manifest.json (WebExtension) find what they're looking for (Test Pilot signing)

And switch to using jpm to build the xpi since it does what we need and we already have it, so no need for a zip dependency

Related to #25 #26 

---

Once this merges I'll attempt another build on `production` and check the Jenkins result

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,10 +37,10 @@
     "pretest": "npm run-script autolint",
     "test": "cross-env NODE_ENV=test nyc mocha-webpack --opts ./mocha-webpack.opts --webpack-config ./webpack.config.test.js -r test/setup.js test",
     "codecov": "nyc report --reporter=lcov && codecov",
-    "clean": "git clean -fX ./dist ./*.xpi && npm run :clean:doc ",
+    "clean": "git clean -fX ./dist ./*.xpi ./addon.env && npm run :clean:doc ",
     "distclean": "git clean -dfX",
     "prepackage": "npm run-script clean && cross-env NODE_ENV=production npm run-script build",
-    "package": "cd dist && nodezip -c ../addon.xpi *"
+    "package": "echo 'TESTPILOT_ADDON_ID='${npm_package_id}'\nTESTPILOT_ADDON_VERSION='${npm_package_version} > addon.env && cd dist && nodezip -c ../addon.xpi *"
   },
   "nyc": {
     "all": true,
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "clean": "git clean -fX ./dist ./*.xpi && npm run :clean:doc ",
     "distclean": "git clean -dfX",
     "prepackage": "npm run-script clean && cross-env NODE_ENV=production npm run-script build",
-    "package": "jpm xpi --addon-dir dist/ --dest-dir . && mv lockbox.xpi addon.xpi"
+    "package": "cd dist && nodezip -c ../addon.xpi *"
   },
   "nyc": {
     "all": true,
@@ -106,6 +106,7 @@
     "json-templater": "^1.1.0",
     "mocha": "^3.5.3",
     "mocha-webpack": "^1.0.0-rc.1",
+    "node-zip": "^1.1.1",
     "nyc": "^11.2.1",
     "raw-loader": "^0.5.1",
     "react-addons-test-utils": "^15.6.0",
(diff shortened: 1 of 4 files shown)


jimporter:
> This way pipeline scripts looking for a package.json (addon) or manifest.json (WebExtension) find what they're looking for (Test Pilot signing)

This sounds broken... are we sure that's supposed to be required?

devinreams:
>This sounds broken... are we sure that's supposed to be required?

Yeah, I'll ask about using values from `install.rdf` instead but it's a home-grown script so this was my quick-n-dirty way to get past the errors and see if we get any closer to successful signing.

As discussed, ideally if we need this (versus a install.rdf check) we create a slimmed-down package.json.

devinreams:
Never mind. Change of plan: we can just write to `addon.env` on package.

```
TESTPILOT_ADDON_ID=@lockbox
TESTPILOT_ADDON_VERSION=WHATEVERVERSIONWEAREON
```

devinreams:
@jimporter How about just this `addon.env` file for now?

Any concerns before I merge and test it against the signing pipeline?

package.json
```diff
@@ -37,10 +37,10 @@
     "pretest": "npm run-script autolint",
     "test": "cross-env NODE_ENV=test nyc mocha-webpack --opts ./mocha-webpack.opts --webpack-config ./webpack.config.test.js -r test/setup.js test",
     "codecov": "nyc report --reporter=lcov && codecov",
-    "clean": "git clean -fX ./dist ./*.xpi && npm run :clean:doc ",
+    "clean": "git clean -fX ./dist ./*.xpi ./addon.env && npm run :clean:doc ",
     "distclean": "git clean -dfX",
     "prepackage": "npm run-script clean && cross-env NODE_ENV=production npm run-script build",
-    "package": "cd dist && nodezip -c ../addon.xpi *"
+    "package": "echo 'TESTPILOT_ADDON_ID='${npm_package_id}'\nTESTPILOT_ADDON_VERSION='${npm_package_version} > addon.env && cd dist && nodezip -c ../addon.xpi *"
```
jimporter: Are the single quotes necessary? I'm pretty sure this will make Windows unhappy.

devinreams: >Are the single quotes necessary? I'm pretty sure this will make Windows unhappy.

Yes. Or at least: I wasn't able to find the right combination of output to both a) get the variables properly output and b) write to the addon.env file.

In other words, without single quotes around _just_ the text strings (sans variables):

```
"package": "echo TESTPILOT_ADDON_ID=${npm_package_id}\nTESTPILOT_ADDON_VERSION=${npm_package_version} > addon.env && cd dist && nodezip -c ../addon.xpi *"```

...nothing actually is written to addon.env in my environment.

And with single quotes around the entire echo:

```
"package": "echo 'TESTPILOT_ADDON_ID=${npm_package_id}\nTESTPILOT_ADDON_VERSION=${npm_package_version}' > addon.env && cd dist && nodezip -c ../addon.xpi *"
```

...the file is written to, but with "${npm_package_id}" and "${npm_package_version}" output  instead of the expected parsed variable ("lockbox@mozilla.com", "0.1.0-pre").

Therefore... I arrived at what you currently see here. 😬 

Any other ideas worth trying right now?

Comment (jimporter) [reply package.json]: I think the newline is causing issues. I would split it into two statements: `echo TESTPILOT_ADDON_ID=${npm_package_id} > addon.env && echo TESTPILOT_ADDON_VERSION=${npm_package_version} >> addon.env`

Comment (pdehaan) [reply package.json]: I had some success with the following:

```sh
"package": "echo \"TESTPILOT_ADDON_ID=$npm_package_id\nTESTPILOT_ADDON_VERSION=$npm_package_version\" > addon.env && cd dist && nodezip -c ../addon.xpi *"
```

That said, this feels funny and dirty to me, and very Mac/UNIX specific and I'm not sure how well it'd perform on Windows. Not sure if it'd be better to just replace with some basic Node shell script:

```sh
#!/usr/bin/env node

const {writeFileSync} = require('fs');
const {id, version} = require('../package.json');

const output = `TESTPILOT_ADDON_ID=${id}\nTESTPILOT_ADDON_VERSION=${version}`;

writeFileSync('./addon.env', output);
```

devinreams: Good suggestions. Thanks, everyone.

I don't have a great way to test this until I actually push to production. So while I agree this feels a bit brittle and O/S specific, but to avoid more pollution or clutter here, I'll take the "good enough" route for now and make sure this builds in the Jenkins env leaving room to improve if it does actually work. 😉

Issue #75, opened 2017-09-12
Pull json-webpack-plugin out into its own npm package


jimporter:
`json-webpack-plugin` might be useful for other Webpack users, so we should turn it into a package and host it on npm. This can wait until after alpha though.

PR #123, opened 2017-10-03
Automatically generate a list of available locales

jimporter:
This is pretty simple; it just adds the necessary boilerplate to make a `locales.json` file with a list of the available locales. This way it'll be easier for people to add new locales down the line.


dir-list-webpack-plugin.js
```diff
@@ -0,0 +1,51 @@
...
+      fs.readdir(this.options.directory, (err, files) => {
+        if (err) {
+          compilation.errors.push(`DirListWebpackPlugin couldn't read ` +
+                                  `directory $(this.options.directory)`);
+          callback();
+          return;
+        }
+
+        // Ignore dotfiles.
+        const output = JSON.stringify(files.filter((x) => x.charAt(0) !== ".");
```
pdehaan: ESLint really doesn't like this line. I think you're missing a closing `)` before the `;`.

webpack.config.babel.js
```diff
@@ -118,10 +119,14 @@ export default {
...
     }]}),
     new JSONWebpackPlugin({
-      template: "src/webextension/manifest.json.tpl",
+      template: path.join(__dirname, "src/webextension/manifest.json.tpl"),
       filename: "webextension/manifest.json",
       data: thisPackage,
     }),
+    new DirListWebpackPlugin({
+      directory: path.join(__dirname, "src/webextension/locales"),
+      filename: "webextension/locales/locales.json",
```
pdehaan: I'm a bit confused here... So, it seems like we're just grabbing a list of all files/folders from a specified directory and outputting it to a "locales.json" file here. That seems OK and we've done that for other add-ons for development, but usually for production we'd check translation %%s and make sure a locale is 95-100% translated in Pontoon before adding it as an available locale. Otherwise some users may get 15% Russian and then 85% fallback English strings (and a poor user experience, or whatever).

dir-list-webpack-plugin.js
```diff
@@ -0,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import fs from "fs";
+
+export default class DirListWebpackPlugin {
```
pdehaan: Similar to _"Pull json-webpack-plugin out into its own npm package #75"_, this plugin seems pretty awesome and probably worthy of extracting out into a dedicated npm module (in the future).

Give a 👍/:-1: and I can file a separate tracking bug for our future selves.

webpack.config.babel.js
jimporter: Should we just manually update locales.json then? Or maybe just do this automatic thingy for dev builds?

pdehaan: I think this is fine for now, but we may need to tweak it once we get closer to a beta release and get wired into pontoon.

jimporter: Well, I added a production config that needs to be manually updated, plus a README explaining how it works for localizers.

dir-list-webpack-plugin.js
```diff
@@ -0,0 +1,81 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import fs from "fs";
```
linuxwolf: Not saying we do it now, but ...

`[promisify-node](https://github.com/nodegit/promisify-node)` helps with these.  I'd only suggest looking more seriously at it if more node packages need promisification.

pdehaan: Also, Node 8 has a built in [`util.promisify`](https://nodejs.org/api/util.html#util_util_promisify_original).

PR #124, opened 2017-10-03
Make our <ItemFilter/> always display the current redux state of the filter

jimporter:
This is a small cleanup patch to ensure that the filter stays in sync with the Redux store.


jimporter:
Force-merged since all I did was rebase.

PR #126, opened 2017-10-03
Add confirmation prompts

### LINKED ISSUE #125: Editor confirmations to prevent data loss
From @jimporter 

- [ ] when updating/editing: to cancel throw a confirmation
- [ ] clicking delete: to actually delete, throw a confirmation

jimporter:
Resolves #125 

Notable bits:
* I updated jsdom to resolve https://github.com/tmpvar/jsdom/issues/1959
* Focus should mostly work, but you lose your current focus after closing a modal dialog (this needs fixed, but I'm not sure how to do it yet)


src/webextension/manage/containers/modal-root.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+ModalRoot.propTypes = {
+  modal: PropTypes.string,
+  getString: PropTypes.func.isRequired,
+};
+
+export default connect(
+  (state) => ({
+    modal: state.modal.modal,
```
linuxwolf: why the two layers (`modal.modal`)?

linuxwolf: I would consider `.name` instead of `.modal`.  Where it's used it should be clear which one is the target.

jimporter: Good point. I went with `.id`, since I think that emphasizes that it's a unique internal name.

jimporter:
@linuxwolf Just to be sure, this is ok to merge now?

jimporter:
Force-pushed after rebasing.

PR #131, opened 2017-10-04
Replace node-zip with jpm for XPI packaging

### LINKED ISSUE #127: Addon signing appears to mangle binary files? (broken icon)
Observed today after downloading a signed version of the extension here:
http://testpilot.firefox.com/files/lockbox@mozilla.com/signed-addon.xpi

The icons don't load.

After investigating, it seems binary files are slightly.. affected? at some point in the Jenkins build/signing pipeline. Perhaps?

A binary diff on the `dist` output and the `signed-addon.xpi` shows some strays appearing, like this `C2` at the beginning of the icon.png file:

screen_shot_2017-10-03_at_7_32_58_pm

If you diff a plain text file (json, html) there's no differences. If you diff another non-plain text file (like our manage.ftl) you see differences.

---

I'll poke around and see who can help troubleshoot this...

devinreams:
Hopefully fixes #127 

@jimporter can you please test and run on your machine and confirm package works and the xpi loads ok for you? works on mine. 😁

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "clean": "git clean -fX ./dist ./*.xpi ./addon.env && npm run :clean:doc ",
     "distclean": "git clean -dfX",
     "prepackage": "npm run-script clean && cross-env NODE_ENV=production npm run-script build",
-    "package": "echo TESTPILOT_ADDON_ID=${npm_package_id} > addon.env && echo TESTPILOT_ADDON_VERSION=${npm_package_version} >> addon.env && cd dist && nodezip -c ../addon.xpi *"
+    "package": "echo TESTPILOT_ADDON_ID=${npm_package_id} > addon.env && echo TESTPILOT_ADDON_VERSION=${npm_package_version} >> addon.env && cd dist && jpm xpi --dest-dir ../ && mv ../null.xpi ../addon.xpi"
   },
   "nyc": {
     "all": true,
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "clean": "git clean -fX ./dist ./*.xpi ./addon.env && npm run :clean:doc ",
     "distclean": "git clean -dfX",
     "prepackage": "npm run-script clean && cross-env NODE_ENV=production npm run-script build",
-    "package": "echo TESTPILOT_ADDON_ID=${npm_package_id} > addon.env && echo TESTPILOT_ADDON_VERSION=${npm_package_version} >> addon.env && cd dist && nodezip -c ../addon.xpi *"
+    "package": "echo TESTPILOT_ADDON_ID=${npm_package_id} > addon.env && echo TESTPILOT_ADDON_VERSION=${npm_package_version} >> addon.env && cd dist && jpm xpi --dest-dir ../ && mv ../null.xpi ../addon.xpi"
```
pdehaan: Instead of `cd dist`, I _think_ we can specify the `--addon-dir`, like this:

```json
jpm xpi --addon-dir=dist/ --dest-dir=./
```

Although the only way I can seemingly get around the odd `null.xpi` filename is if I copy the package.json file into dist/package.json.

devinreams: >I think we can specify the --addon-dir

Good catch. 👍

devinreams:
I threw `--addon-dir` to avoid the `cd` but kept the `mv` to avoid more file shuffling and keep the packaging steps super obvious here...

devinreams:
Tested package and the resulting xpi still works as expected.

I'm just waiting on a quick fix for `locales.json` to merge first.

Then I will merge this and merge into "production" branch and see what the signing pipeline does for us this time...

PR #132, opened 2017-10-04
Fix locales

jimporter:


jimporter:
Force-pushed after just a rebase.

PR #139, opened 2017-10-05
Bump version to 0.1.0-pre2

devinreams:
It turns out node will interpret a version of "0.1.0pre2" and return it as "0.1.0-pre2". Whereas our templates take the literal string.

So the addon.env has the hyphenated version (which goes to update.json) but the manifest has the non-hyphenated version. So, let's just go all hyphen (semver?) all the time! 🎉 

Once this is live and signed, I still need to test and make sure this goes out to updates.json and the browser properly does a version comparison (0.1.0pre1 is what Firefox has installed, and updates.json will say there is a 0.1.0-pre2, hopefully the latter is greater than the former).


devinreams:
@jimporter are you OK with me merging this and deploying to `production` to test the build/sign/deploy pipeline again?

devinreams:
> I'll probably patch our install.rdf to convert to AMO-style versions.

Ah, nice idea. 

I wonder if being inconsistent will cause comparison checks to fail somewhere in some process, though? 😬

devinreams:
Follow-up: I'm running `0.1.0pre1` in a browser and "Check for updates" does not seem to find an update `0.1.0-pre2`. I'll wait an see. But it may be that it's not able to properly compare 0pre1 and 0 (-pre2). 🤷‍♂️

pdehaan:
Not sure if we'd need to ping some AMO people or something, but does "pre" need to be specific "alpha", "beta", "\<other\>" strings?

I only found this bit in the npm semver docs: https://docs.npmjs.com/misc/semver#prerelease-tags
I can try some ~~Googling~~ Yahoo!ing and see if MDN has any similar semver docs, or if we can switch to "0.1.0-alpha.2".

devinreams:
I believe we should be fine with either approach: https://developer.mozilla.org/en-US/Add-ons/Distribution#Beta_versions

It's just that the inconsistency between node's interpretation of a version (turning it into semver) may be an issue. So aiming for semver to match node's expectation _everywhere_ seems best, even if the Toolkit version spec says we should only separate by dots.

And the fact I have a non-semver version (0.1.0pre1) trying to compare to a semver version (0.1.0-pre2). My hope is I can just install pre2, bump to pre3 and the comparison will sort itself out.

---

Obviously we can avoid this all together and just say Alpha and up is 0.1.* and not use pre/alpha/beta at all. 😄

PR #142, opened 2017-10-05
Style and test FRUX

jimporter:
~WIP~


src/webextension/l10n.js
```diff
@@ -23,24 +23,33 @@ async function fetchMessages(baseDir, locale, bundle) {
...
   }
 }
 
-async function createMessagesGenerator(baseDir, currentLocales, bundle) {
-  const fetched = await Promise.all(
-    currentLocales.map((x) => fetchMessages(baseDir, x, bundle))
-  );
-  const bundles = fetched.reduce(
+async function fetchAllMessages(baseDir, locale, bundles) {
+  return Promise.all(bundles.map((i) => fetchMessages(baseDir, locale, i)));
```
pdehaan: Does this need an `await`?

jimporter: No, but the `async` is probably superfluous now.

jimporter:
@linuxwolf The only changes I made after the rebase were to add `browser.storage.local` mocks and to pretend that we're signed in when we initialize the datastore. I do that via:

```js
    // Pretend we're signed in.
    getAuthorization().info = {};
```

src/webextension/firstrun/components/wizard-page.js
```diff
@@ -0,0 +1,35 @@
...
+
+import Button from "../../widgets/button";
+
+export default class WizardPage extends React.Component {
+  static get propTypes() {
+    return {
+      title: PropTypes.string.isRequired,
+      submitLabel: PropTypes.string.isRequired,
+      onSubmit: PropTypes.func.isRequired,
+      children: PropTypes.any,
```
pdehaan: nit: Should this be `PropTypes.node`?

> _"Generally, you should use `PropTypes.node` for `children`. It accepts anything that can be rendered: numbers, strings, elements or an array containing these types."_
> 
> &mdash; via https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md#as-for-exceptions

jimporter:
Merging; all I did was replace `PropTypes.any` with `PropTypes.node`.

PR #143, opened 2017-10-05
Save (and Load) some User Info

### LINKED ISSUE #135: Capture and store FxA UID
From the first run flow, capture and store the Firefox Accounts user ID.

Then we can pass it as an "extra" value to Telemetry #118

linuxwolf:
This saves and loads the FxA user id; in preparation for work on telemetry.

Fixes #135


linuxwolf:
merge override reason: only change is rebase to master.

PR #144, opened 2017-10-06
Add telemetry

jimporter:
(Only the last commit is relevant.)

@linuxwolf This is a rough sketch of what I'm thinking about for Telemetry. The goal is to have a simple API that looks like the XPCOM Telemetry API, since that should hopefully make it easy to convert to WebExtension-based Telemetry if/when we get it.

Some extra stuff that might be worth doing:
* Defining our events in a separate JSON file?
* Making a similar API for our background script events? This might make sender-side error-detection easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "babel-loader": "^7.1.2",
     "babel-minify-webpack-plugin": "^0.2.0",
     "babel-plugin-istanbul": "^4.1.5",
+    "babel-plugin-rewire": "^1.1.0",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-polyfill": "^6.23.0",
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -4,9 +4,36 @@
 
 /* eslint-disable no-unused-vars */
 
+const { utils: Cu } = Components;
+
+Cu.import("resource://gre/modules/Services.jsm");
+
 function startup({webExtension}, reason) {
-  webExtension.startup().then(() => {
+  Services.telemetry.registerEvents("lockbox", {
+    "startup": {
+      methods: ["startup"],
+      objects: ["addon", "webextension"],
+    },
+    "click": {
+      methods: ["click"],
+      objects: ["browser_action"],
+      extra_keys: ["fxauid"],
+    },
+  });
+
+  webExtension.startup().then(({browser}) => {
     console.log("embedded webextension has started");
+    Services.telemetry.recordEvent("lockbox", "startup", "webextension");
+    browser.runtime.onMessage.addListener((message, sender, respond) => {
+      switch (message.type) {
+      case "telemetry_event":
+        Services.telemetry.recordEvent(
+          message.category, message.method, message.object, null,
+          message.extra || null
+        );
+        respond({});
+      }
+    });
   });
 }
 
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -40,3 +40,10 @@ function startup({webExtension}, reason) {
 function shutdown(data, reason) {}
 function install(data, reason) {}
 function uninstall(data, reason) {}
+
+// We need to reference these functions so that babel-plugin-rewire can see
+// them for our tests.
+startup;
+shutdown;
+install;
+uninstall;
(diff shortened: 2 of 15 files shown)


src/webextension/telemetry.js
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export async function recordEvent(category, method, object, value, extra) {
```
linuxwolf: There's a few pieces of common data we'll need to gather for just about every event:

* FxA user identifier, when available (gotten from `Authorization.uid`)
* Item identifier, when available (gotten from `item.id`)
* action/flow identifier (preferably a UUID so it has a good chance of being unique across runs, views)

src/bootstrap.js
```diff
@@ -4,9 +4,35 @@
...
+    },
+    "click": {
+      methods: ["click"],
+      objects: ["browser_action"],
+    },
+  });
+
+  webExtension.startup().then(({browser}) => {
     console.log("embedded webextension has started");
+    Services.telemetry.recordEvent("lockbox", "startup", "webextension");
```
linuxwolf: (cc: @irrationalagent )

Do you want the user identifier (if available) on startup?

linuxwolf: Assuming the uid is desired, I think this needs to be ported into `webextension/background/index.js`.

src/bootstrap.js
```diff
@@ -4,9 +4,35 @@
 
 /* eslint-disable no-unused-vars */
 
+const { utils: Cu } = Components;
+
+Cu.import("resource://gre/modules/Services.jsm");
+
 function startup({webExtension}, reason) {
-  webExtension.startup().then(() => {
+  Services.telemetry.registerEvents("lockbox", {
```
linuxwolf: I don't know if an external JSON file is worthwhile.  Each event will involve corresponding code changes to actually record the event.

jimporter:
@linuxwolf How does this look? Note that I used `fxauid` instead of `fxa_uid` because the telemetry API doesn't like the latter.

linuxwolf:
@irrationalagent note the comment above from @jimporter -- "fxa_uid" was not acceptable to `Services.telemetry`, but "fxauid" was acceptable.

src/webextension/telemetry.js
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
linuxwolf: Is this file still used?

src/webextension/background/telemetry.js
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import getAuthorization from "./authorization/index";
+
+export async function recordEvent(category, method, object, value, extra) {
```
linuxwolf: The current metrics spec does not use `value`, so I think it's fine to omit it.

src/webextension/telemetry.js
jimporter: Not currently, but it will be for any telemetry events recorded in content pages (e.g. the management UI).

linuxwolf:
@devinreams 

We did discuss breaking things down.  I'm starting on work to support "lockbox_datastore." metrics, while @jimporter was going to tackle the front-end items based on the priorities guidelines in metrics.md.

irrationalagent:
is there a plan to merge this soon (or create a shared branch) or if I want to add to these changes should i make a branch off of jim's?

linuxwolf:
@jimporter can you migrate this to a shared branch in this repo soon?  I, too, have updates I'd like to telemetry work.

jimporter:
We should probably just put it on master, really. Aside from review comments and tests, this should be good to land for real.

linuxwolf:
@jimporter That's fine.

:shipit:

devinreams:
>We should probably just put it on master, really. 

Yeah! 👍 🌊🚢

Then @irrationalagent can push what he works on today and we can pick up and run with that branch...

jimporter:
@linuxwolf Ok, this is ready for another review. I added a bunch of tests here so coverage should be pretty good.

package.json
```diff
@@ -74,6 +74,7 @@
     "babel-loader": "^7.1.2",
     "babel-minify-webpack-plugin": "^0.2.0",
     "babel-plugin-istanbul": "^4.1.5",
+    "babel-plugin-rewire": "^1.1.0",
```
linuxwolf: interesting find

devinreams:
I just added Lockbox (0.1.0-pre3) to my primary browser and went through the firstrun flow to collect Telemetry.

screen shot 2017-10-11 at 10 19 35 am

@irrationalagent please check the servers and confirm this is actually arriving. 😁

PR #145, opened 2017-10-06
Ignore coverage to prevent CI warnings on report

devinreams:
Builds on tags run `npm run package` after codecov has run (to build the XPI and add it to GitHub Releases).

But package runs eslint. And eslint then runs against the "coverage" directory?

https://travis-ci.org/mozilla-lockbox/lockbox-extension/builds/284306037#L4257

Would this change to ignore prevent all those warnings? ☝️

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,6 @@
 !.eslintrc.js
+coverage
 dist
-site
 node_modules
+site
 temp


linuxwolf:
This change should prevent `eslint` from looking into `/coverage`.

jimporter:
The Travis warnings are because of all the console.logs. I'll file an issue to discuss what our console policy should be.

devinreams:
Well yeah, there are _those_. But, I'm looking at the ones that happen at "before_deploy" when it hits lcov's "prettify.js":

screen shot 2017-10-06 at 6 13 52 pm

..perhaps can also be remedied by us cleaning and removing the "coverage" directory.

jimporter:
Ah ha. Yeah, that's a problem.

PR #146, opened 2017-10-06
Move updateURL to install.rdf so updates are actually checked

devinreams:
Long story short, @jimporter was right: https://github.com/mozilla-lockbox/lockbox-extension/pull/62#pullrequestreview-64110542

We definitely need the `updates.json` to live in install.rdf for our non-WebExtension extension. Whoops.

Tested and confirmed locally that the updates screen does attempt to check the remote `updates.json` whereas it is definitely _not_ doing that without this change.

screen shot 2017-10-06 at 2 26 02 pm

Sorry for the extra noise/hassle here. 😔

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "lockbox",
-  "version": "0.1.0-pre2",
+  "version": "0.1.0-pre3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,23 +55,6 @@
         }
       }
     },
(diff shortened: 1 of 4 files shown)


devinreams:
>We should probably remove this from manifest.json as well, since it's currently a no-op there.

Agreed... 

Removed and confirmed `updates.json` is _still_ requested in the add-ons update. _Now_ maybe this thing will update itself. 😏 

screen shot 2017-10-07 at 12 05 45 pm

devinreams:
After we land a few things I'd like to merge this and bump the version so our next update (hopefully) gets picked up in the browser.

@jimporter Any other concerns or feedback on this PR?

PR #147, opened 2017-10-09
Replace icon and optimize for multiple states via svg implementation

changecourse:
Updated icons for locked/unlocked state of lockbox, now in SVG!

- [x] Replace png in WebExtension manifest (for toolbar)
  - [x] Add toolbar logic to support different states (locked and unlocked)
- [x] Replace png favicon for svg (not [widely supported](http://caniuse.com/#feat=link-icon-svg), but fine for us)
- [x] Replace png icon for `about:addons` (need an icon.png in top-level dist)


devinreams:
Thanks @changecourse. I'll help with this ✋ stay tuned...

devinreams:
Well, I didn't get this as far as I'd like:

- Turns out I couldn't quickly determine how to properly change the icon on unlock status. It's obviously [browserAction.setIcon](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/browserAction/setIcon) but I wasn't sure where to hook into our messaging.

- Turns out the easiest way to get an icon on the `config:addons` screen for a legacy addon is to just drop an icon.png in the top level directory. Otherwise you have to define it in install.rdf's `em:iconURL` which I don't know how to programmatically get the path to the WebExtension SVG to. (SVG does work, just can't figure out the path - is it worth fighting this or just get a PNG and move it using webpack?)

@jimporter Is this something you're OK with picking up and helping with from here?

devinreams:
@changecourse do you consider these changes a "must have" for Alpha?

I ask because if the two items left ☝️ aren't super straight forward perhaps we split those out into a separate issue to tackle afterwards and get the new icon as-is landed first?

changecourse:
I think that's fine... it would be nice to have the lock/unlock indication and be able to even see if the first pass even makes sense/is noticeable for our alpha users, but it's not a deal-breaker either.

If that's the case, we should move forward just using the locked icon for everything, in lieu of not having state changes.

linuxwolf:
@devinreams 
> * Turns out I couldn't quickly determine how to properly change the icon on unlock status. It's obviously browserAction.setIcon but I wasn't sure where to hook into our messaging.

[browser-action.js](https://github.com/mozilla-lockbox/lockbox-extension/blob/3167bb210fff2d9921a4aff6d59482521c939422/src/webextension/background/browser-action.js#L34) has a place it looks at the datastore's state, and is (so far) called when the datastore global state changes.

> * Turns out the easiest way to get an icon on the `config:addons` screen for a legacy addon is to just drop an icon.png in the top level directory. Otherwise you have to define it in install.rdf's `em:iconURL` which I don't know how to programmatically get the path to the WebExtension SVG to. (SVG does work, just can't figure out the path - is it worth fighting this or just get a PNG and move it using webpack?)

I guess I don't see why we can't have a PNG for where the legacy addon needs it (presumably just in `about:addons`), and the SVG where the webextension needs that (presumably everywhere else).  If that's acceptable, sourcing the SVG within the webextension and calling `browser.extension.getURL("icons/${state}.svg")` should work ... in theory (-;

jimporter:
Yeah, let's just have a separate icon file for the add-on's image. That should be the easiest.

devinreams:
>Yeah, let's just have a separate icon file for the add-on's image. That should be the easiest.

Ok 😅 that's what I wanted to do, too. Was probably just over-engineering the use of the svg.

@irrationalagent Can you please generate a PNG that's 96px and drop it back in to `webextension/icons/lock.png` so we can use that on the addons page?

>browser-action.js has a place it looks at the datastore's state, and is (so far) called when the datastore global state changes.

Ah yes, I'll poke around a little more and see if I can figure that out...

changecourse:
you said @irrationalagent but I think you meant me ;)

added png to folder.

devinreams:
>added png to folder.

Thanks. ✅ 

That's wired back up:

screen shot 2017-10-09 at 2 34 54 pm

Along with the browser action:

screen shot 2017-10-09 at 2 35 22 pm


Last up: dynamically change the toolbar icon if unlocked...

devinreams:
Well I think I have it switching to "unlock" icon at the right time...

lock-unlock

Can someone tell me if this is an appropriate place to hook on to, too? I tested that this works as expected even at first run (unlocked does not ever appear, from what I tested, even though a popup doesn't exist yet).

--

@changecourse any other icon details you want to tweak while we're in here?

src/webextension/background/browser-action.js
```diff
@@ -12,10 +12,12 @@ function installPopup(path) {
   browser.browserAction.setPopup({
     popup,
   });
+  browser.browserAction.setIcon({path: "icons/lb_locked.svg"});
```
linuxwolf: Setting the icon would be better done in `updateBrowserAction()` than here.

jimporter: This should probably go down in `updateBrowserAction`.

src/webextension/background/browser-action.js
```diff
@@ -12,10 +12,12 @@ function installPopup(path) {
...
 }
 function uninstallPopup() {
   if (popup) {
     browser.browserAction.setPopup({ popup: "" });
+    browser.browserAction.setIcon({path: "icons/lb_unlocked.svg"});
```
jimporter: Ditto.

webpack.config.babel.js
```diff
@@ -135,7 +135,7 @@ export default {
       inject: false,
       minify: htmlMinifyOptions,
       title: "Welcome to Lockbox",
-      icon: "../icons/lock.png",
+      icon: "../icons/lb_locked.svg",
```
jimporter: We could probably remove the icon from here entirely; it's just a popup.

webpack.config.babel.js
```diff
@@ -144,7 +144,7 @@ export default {
       inject: false,
       minify: htmlMinifyOptions,
       title: "Unlock",
-      icon: "../icons/lock.png",
+      icon: "../icons/lb_locked.svg",
```
jimporter: Ditto.

webpack.config.babel.js
```diff
@@ -126,7 +126,7 @@ export default {
       inject: false,
       minify: htmlMinifyOptions,
       title: "Lockbox Entries",
-      icon: "../icons/lock.png",
+      icon: "../icons/lb_locked.svg",
```
jimporter: @changecourse Should this be the locked or unlocked icon? You can only see the entries when you're unlocked, but the locked icon is the main logo.

devinreams:
Good feedback @jimporter. @linuxwolf also just suggested a ternary to set the path. Thanks for the help. Updates incoming...

With the tests failing, do I need a case in the listener? I can't quite decypher the errors/code. 😬

webpack.config.babel.js
devinreams: >Should this be the locked or unlocked icon? You can only see the entries when you're unlocked, but the locked icon is the main logo.

@jimporter @changecourse I went ahead and changed the favicon to the unlocked icon for consistency. It's easy enough to switch back or replace entirely if we want...

devinreams:
@jimporter @linuxwolf @changecourse OK this _should_ be good to go now.

Can you please review again and double check for me?

1. updated favicon to point to SVG
2. updated browser action icon to SVG
3. made browser action icon dynamic based on lock/unlock state
4. re-added a simple PNG icon for legacy addon settings page

Thank you 😁 

## Unlocked

screen shot 2017-10-10 at 9 25 50 am

## Locked

screen shot 2017-10-10 at 9 26 36 am

src/template.ejs
```diff
@@ -14,7 +14,7 @@
     <% } for (let js of htmlWebpackPlugin.files.js) { %>
     <script src="<%= path.relative(dir, js) %>" defer></script>
     <% } %>
-    <link href="<%= htmlWebpackPlugin.options.icon %>" rel="shortcut icon" type="image/png"/>
+    <link href="<%= htmlWebpackPlugin.options.icon %>" rel="shortcut icon" type="image/svg"/>
```
jimporter: Hmm, I don't like that the mimetype is hardcoded here, but that's my fault, so I'll just file an issue about it and fix it in a followup... :(

src/webextension/background/browser-action.js
```diff
@@ -37,6 +37,9 @@ export default async function updateBrowserAction(ds) {
   uninstallListener();
   uninstallPopup();
 
+  let iconpath = ds.locked ? "icons/lb_locked.svg" : "icons/lb_unlocked.svg";
```
jimporter: `s/let/const/` please. I've been trying to be as "const-correct" as possible, though it matters significantly less here than in C++.

devinreams:
>`s/let/const/` please.

@jimporter Got it. Thank you! 😀

PR #148, opened 2017-10-09
Settings page for "Factory Reset"

### LINKED ISSUE #137: User interface to reset/clean database
To re-install, wipe data, get out of a broken situation during testing?, etc.

linuxwolf:
Fixes #137 

Adds a Settings/Options page to the extension, that can be used without the user unlocking.

The following still needs to be done:

* [x] implement a confirmation modal
* [x] write tests
* ~reset FxA info (maybe?)~


src/webextension/background/views.js
```diff
@@ -91,6 +91,15 @@ export async function openView(name) {
 }
 
 export async function closeView(name) {
+  if (!name) {
+    console.log("closing all views");
+    Object.keys(views).forEach((v) => {
+      v = views[v];
+      v.close();
+    });
```
pdehaan: Not sure how you feel about recursion, but this shaves a few chars (and keeps the close view logic in one less place):
```js
Object.keys(views).forEach(v => closeView(v));
```

linuxwolf: With the convention we're following here, it would be:

```javascript
Object.keys(views).forEach((v) => closeView(v));
```

But I don't have a strong opinion either way.

devinreams:
Mentioned today, would be nice to have a modal dialog to confirm in place... @jimporter do you have a moment to help?

test/settings/components/app-test.js
```diff
@@ -0,0 +1,23 @@
...
+
+import chai, { expect } from "chai";
+import chaiEnzyme from "chai-enzyme";
+import React from "react";
+
+chai.use(chaiEnzyme());
+
+import mountWithL10n from "../../mock-l10n";
+import App from "../../../src/webextension/settings/components/app";
+import FactoryReset from "../../../src/webextension/settings/components/factory-reset";
```
jimporter: We have nice aliases for these paths now: `src/foo/bar` and `test/foo/bar`

test/settings/components/factory-reset-test.js
```diff
@@ -0,0 +1,31 @@
...
+require("babel-polyfill");
+
+import chai, { expect } from "chai";
+import chaiEnzyme from "chai-enzyme";
+import React from "react";
+
+chai.use(chaiEnzyme);
+
+import mountWithL10n from "../../mock-l10n";
+import FactoryReset from "../../../src/webextension/settings/components/factory-reset";
```
jimporter: As above, we have aliases for the import paths now.

linuxwolf:
This was ready a while ago, but forgot this branch was on the shared repo \-:

@jimporter should be ready now

src/webextension/background/views.js
```diff
@@ -63,6 +63,15 @@ export async function openView(name) {
 }
 
 export async function closeView(name) {
+  if (!name) {
+    console.log("closing all views");
```
jimporter: Since I just removed a bunch of our extra `console.log`s, we probably don't need this anymore. Also, I think I introduced a merge conflict here. Sorry!

src/webextension/settings/actions.js
```diff
@@ -0,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jimporter: Strictly speaking, it's not necessary for the settings page to use Redux if you don't want. I've been hemming and hawing about where we should use Redux (aside from the management UI, where we really need it), so I'm open to suggestions.

In the longer term, if we do go with Redux here, we might want to pull out the Redux modal bits into a common place so that we're not duplicating as much code.

linuxwolf: Long-term, I'm not sure either.  For what this does _today_, it's kind of overkill.  For what I think we'll want in the future (automatic sign out options, filling behaviors, etc), it might still be needed.

If all of that necessitates Redux (which I suspect it will) then it felt like it would be more work to move things around later if we wanted more states than to just be ready for it now, hence the structure decisions I made.

Issue #151, opened 2017-10-09
Only allow one editor view of lockbox open at a time


changecourse:
After doing some testing with the current build, as well as trying to evaluate how the lock/unlock functionality will work (both in the popover as well as the tab) I worry that it will create some confusion with having multiple versions of the entry list open at a time, each in potentially different states... If the user locks lockbox in one tab, it still requires a refresh in the other tabs to take effect, and the user is left to wonder whether or not the entries are still accessible in a 'locked' state.

jimporter:
We can lock background tabs without any issues. I vastly prefer allowing multiple management tabs, since it makes it possible to view one item while editing another (e.g. to copy notes across).

changecourse:
I hear you about copying/pasting between different entries, and yes, allowing multiple tabs to be open would help in this, but I don't think that use case outweighs the confusion or potential for concern around visibility of their lockbox (even if we lock it).

We're making some big assumptions about...
1. someone using the notes field for an entry (it's not a core part of the username/password/url equation)
2. someone also wanting to copy and paste things like notes across multiple entries

I think we should default to one view, not risk user confusion around multiple tabs, and then check analytics around alpha usage to inform some of the use cases you're concerned about for the future.

jimporter:
> I think we should default to one view, not risk user confusion around multiple tabs, and then check analytics around alpha usage to inform some of the use cases you're concerned about for the future.

If we're doing analytics for this, wouldn't it be easier for now if we allowed multiple views so we could track if people actually used them? Switching over to a single view should be easy in the future, since we already have code to do that for first-run.

changecourse:
I would need to understand what we'd be tracking... numbers around how many instances of "multiple tabs open" may not be valuable, because someone could have done that by accident, in its current state. 

Would we track instances of multiple tabs open where fields were copied and pasted between tabs? 

I think this all comes down to hypotheses and which is more important to ensure... the potential for compromising a user's trust/understanding of the security of their lockbox, or the potential for users to copy/paste in a slightly more convenient way between entries (which I still don't believe to be a common use case)

jimporter:
We could track when tabs are focused/unfocused (or when they're part of different windows) and what actions people are performing in each. From there, we could probably get a pretty good idea if people are using multiple editors at once. We could also just ask for feedback on it, since this is pretty early on.

[...]

PR #153, opened 2017-10-09
Start a very simple lock button

### LINKED ISSUE #149: Allow user to "Lock" editor via "Sign Out" to prevent unwanted access to Lockbox data
When letting a friend borrow my computer and browser, I need to be confident that the data in Lockbox is secure-able and control-able by me, so that I can prevent unauthorized access to my credentials locally.

For Alpha, this involves the option to "Sign Out" similar to other FxA reliers, as seen in this visual design:  https://mozilla.invisionapp.com/share/MHEBCD4PK#/screens/264252691

## Acceptance
- When FxA is attached to Lockbox account, there is an option "sign out"
- Datastore will always be "unlocked" and local for users that don't have FxA  attached (just like Firefox today without a master password)

## ToDo
- [x] implement "light" sign out vs. "full" sign out
- [x] implement "signout" backend message
- [ ] implement user action (#442 )

## Future considerations
- Research questions specifically as it relates to locking via sign out & terminology research (what words do we find our target audience using?)
- Lock / Unlock visual interactions to accompany sign out / sign in
- Independent lock/unlock options from sign out/in should target audience require it

devinreams:
Fixes #149

A very quick solution to let the user manually "Lock" or "Sign Out" requiring the user to enter their password again.

I also threw in a reload on the current view just so the user isn't confused why they can still see their passwords. Not great but a start.

lock-it

Some of the concern at #151 (multiple sessions/tabs) comes into play here (user has 3 tabs open, sees items still in two of them, loses trust their data is secure/locked). I'd rather _not_ start any of that decision making/code before Alpha and instead get this work-able and trust-able.

### DIFF
--- a/src/webextension/locales/en-US/manage.ftl
+++ b/src/webextension/locales/en-US/manage.ftl
@@ -1,5 +1,6 @@
 add-item = Add Item
 go-home = Home
+lock-it = Sign Out
 
 item-filter.placeholder = Filter…
 
--- a/src/webextension/manage/components/app.js
+++ b/src/webextension/manage/components/app.js
@@ -8,6 +8,7 @@ import AddItem from "../containers/add-item";
 import AllItems from "../containers/all-items";
 import CurrentSelection from "../containers/current-selection";
 import GoHome from "../containers/go-home";
+import LockIt from "../containers/lock-it";
 import ItemFilter from "../containers/item-filter";
 import ModalRoot from "../containers/modal-root";
 
@@ -25,6 +26,7 @@ export default function App() {
           <menu>
             <AddItem/>
             <GoHome/>
+            <LockIt/>
           </menu>
           <CurrentSelection/>
         </article>
--- a/src/webextension/manage/containers/lock-it.js
+++ b/src/webextension/manage/containers/lock-it.js
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Localized } from "fluent-react";
+import PropTypes from "prop-types";
+import React from "react";
+import { connect } from "react-redux";
+
+import Button from "../../widgets/button";
+import { selectItem } from "../actions";
+
+function LockIt({dispatch}) {
+  return (
+    <Localized id="lock-it">
+      <Button onClick={() => { browser.runtime.sendMessage({type: "lock"});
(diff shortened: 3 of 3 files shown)


jimporter:
"Sign out" seems like a weird name for this; I'd expect that to forget the user's FxA account. Also, we really need to have a "this store is locked" UI for when this happens.

devinreams:
>"Sign out" seems like a weird name for this; I'd expect that to forget the user's FxA account. 

Yeah, discussed a bit here in the design specs provided this morning by @changecourse:

>The act of Locking and Signing Out are very close (in terms of functionality) no?

>I worry that someone may not understand the concept of locking/unlocking at first glance, but that they'd have a sense as to what signing out would do in terms of effectively 'locking' their account access in that manner

And with @linuxwolf response:

>"sign out" and "lock" are probably similar enough to be interchangeable, but I'm not sure. I think it depends on how strongly "sign out" implies "cannot talk to the server", versus "lock" being too technically correct for our target audience to understand.

devinreams:
>Also, we really need to have a "this store is locked" UI for when this happens.

@jimporter Agreed. Do you recommend borrowing the full screen modals as a good starting place perhaps?

That said, I don't know if we should sign up to get to that now before Alpha, especially without UX input and other priorities piling up. Is it cool if I spin off a separate issue?

jimporter:
> That said, I don't know if we should sign up to get to that now before Alpha, especially without UX input and other priorities piling up. Is it cool if I spin off a separate issue?

I really think we need both this and a real "locked" UI, or neither. The current UI is pretty confusing to me.

jimporter:
> Yeah, discussed a bit here in the design specs provided this morning...

Not to be a pain in the butt, but do we have any evidence that "Lock Account" is too confusing? I would imagine users would see that and think along the lines of "Lock" in the Windows Start Menu or the Lock/Unlock terminology used on mobile (I assume macOS has similar terminology). To me, "Sign Out" signifies that we're going to - at minimum - discard all ephemeral state like any Lockbox tabs you have open or entries you're editing, and we don't need to do that. We could have it just override the UI temporarily.

devinreams:
>I really think we need both this and a real "locked" UI. The current UI is pretty confusing to me.

Yeah, I hear you and totally agree. Given the open questions, I propose we first focus on the analytics and UI updates for Alpha. Then prioritize this for into one of the next milestones...

devinreams:
Closing this out as it likely won't be the right approach.

How about we discuss requirements and hold conversation in the issue https://github.com/mozilla-lockbox/lockbox-extension/issues/149 once we're ready to take this on?

Issue #138, opened 2017-10-04
Update user interfaces to match designs


devinreams:
Fonts, styles, spacing, labels, colors, etc.

To be provided...

devinreams:
@jimporter the updated editor designs have been provided by @changecourse in Invision:
https://mozilla.invisionapp.com/share/5JDP20NYB#/screens/256782990

![Screen Shot 2017-10-05 at 9.43.29 AM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/57002eaed1bf020e0086d809/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c57e0b76236454f66b7191a371f116284f5d565edd3c4a4356bcbbd671f3e57862804e0ee485690b7cc53e15ffecd418bd01ba228e4ad1e5eadb1ad2f42f67b6d5db1063474f4dcbed5727a19f41c95df64ce34c5da6123f63e1772431.png)

- assuming still inheriting browser styles as much as possible (buttons don't need to change to be pixel perfect or anything)
- also assuming we "borrow" the Firefox Accounts markup/styles for the first run experience (just for now, not long term, just to help visually bridge the user jumping from Lockbox tab to FxA Popup back to Lockbox)

Let me know if/how you want to split this up, I'm happy to help where I can.

pdehaan:
I'm a bit curious about those icons.
I assume they're just a black&white version of the site's favicon (which should be "easy" enough to [[de]colorize w/ CSS](https://davidwalsh.name/convert-images-grayscale)). But what do we do about sites without favicons, or how would it look with non-transparent favicons?

We have https://github.com/mozilla/tippy-top-sites, which we used for Activity Stream's favicons. Not sure if it'd be useful here. It's still just a random mix of "top sites" on the internet, so wouldn't work for lots of non en_US sites (or even Canadian versions of popular US sites).

Also, would we need to build some sort of image proxy service, so we aren't loading site favicons directly and exposing any referer information?

jimporter:
I think it's black and white as a way of indicating that it's a mockup. My plan is to store favicons in the datastore itself (but not anytime especially soon); this runs the risk of people having outdated icons, but it would also let people customize their icons, which I think would be pretty useful as a way of quickly disambiguating multiple accounts on one site

devinreams:
[...]

PR #154, opened 2017-10-10
Update UX

### LINKED ISSUE #130: Add "send feedback" button to extension

jimporter:
This updates the UX to get us closer to #138. I didn't change the visuals or update strings though; this is just updating the rough layout.

Also fixes #130


package.json
```diff
@@ -39,7 +39,7 @@
     "codecov": "nyc report --reporter=lcov && codecov",
     "clean": "git clean -fX ./dist ./*.xpi ./addon.env && npm run :clean:doc ",
     "distclean": "git clean -dfX",
-    "prepackage": "npm run-script clean && cross-env NODE_ENV=production npm run-script build",
+    "prepackage": "npm run-script clean && npm run-script build",
```
devinreams: If this gets moved out to the `travis.yml`, then I believe the Jenkins CI pipeline that does the extension build and signing (on production branch) won't get set to the right environment. 😔

src/webextension/locales/en-US/manage.ftl
```diff
@@ -1,5 +1,11 @@
-add-item = Add Item
```
devinreams: @jimporter Would you like me to go through and inventory the strings and change them to what @changecourse has proposed?

If so: I can do it here or in my own branch to merge into this one. Your call. 😁

package.json
jimporter: Where's the Jenkins script? Can't we update that?

devinreams: >Where's the Jenkins script? Can't we update that?
No, I don't think we can/should.

It's owned and managed by CloudOps (private mozilla-services repo) and [reused across other Test Pilot projects](https://github.com/mozilla/testpilot/blob/master/docs/development/hosting.md).

Since others rely on what's already there and assumptions are made around it, I don't think we should try to suggest any changes and instead keep making sure our "package" script lands what is expected.

jimporter: This deploy script seems... bad. But I guess it's just one of those things we have to live with.

Still, we should have a way to make a dev XPI so that it doesn't take a minute or more to get a  working XPI for testing. One option would be to make a `dev-package` script; another would be to ensure that the `production` branch always uses `NODE_ENV=production`.

devinreams: >Still, we should have a way to make a dev XPI so that it doesn't take a minute or more to get a working XPI for testing. 

Completely agreed. I figured that's what you were getting at here but wasn't sure.

However you think is best... I'm personally not sure how to approach the latter (tell `production` branch to use the right env) so former may be easiest for now? 😬

src/webextension/locales/en-US/manage.ftl
devinreams: With your hard work on getting everything localized, it turns out it's really easy to update strings. I went through the invision docs and here's a branch in case you want to use it: https://github.com/mozilla-lockbox/lockbox-extension/commits/update-ux-strings

(we can do a PR into this PR if we want to get all complicated about it)

src/webextension/manage/components/send-feedback.js
```diff
@@ -0,0 +1,20 @@
...
+
+import { Localized } from "fluent-react";
+import React from "react";
+
+import Button from "../../widgets/button";
+
+export default function SendFeedback() {
+  return (
+    <Localized id="toolbar-send-feedback">
+      <Button onClick={() => { alert("Sorry, no feedback yet"); }}>
```
devinreams: @jimporter while you're in here, can you please swap out the alert and open this URL (however is best, I'm not sure here) provided by @sandysage: https://qsurvey.mozilla.com/s3/Lockbox-Input

PR #163, opened 2017-10-11
Update UX strings

devinreams:
Fast follow work from #154 related to #138 

~~(will rebase to get this cleaned up)~~

### DIFF
--- a/src/webextension/locales/en-US/manage.ftl
+++ b/src/webextension/locales/en-US/manage.ftl
@@ -7,7 +7,7 @@ toolbar-add-item = New Entry
 toolbar-go-home = Home
 toolbar-send-feedback = Feedback
 
-item-filter.placeholder = Search for an Entry
+item-filter.placeholder = Search for an entry
 
 item-summary-new-item = New Entry
 item-summary-no-title = (No Entry Name)
@@ -16,27 +16,27 @@ item-summary-no-username = (No Username)
 homepage-no-passwords =
   Welcome to Lockbox! I'm Lockie, and I'm here to help you lock
   up your passwords!
-
-  To get started, click { add-item } above.
+  
+  To get started, click { toolbar-add-item } above.
 
 homepage-under-10-passwords = { $count ->
    [1]     Hi there! It's your friend, Lockie!
            
-           You've added { $count } password. That's a great start!
+           You've added { $count } entry. That's a great start!
   *[other] Hi there! It's your friend, Lockie!
            
-           You've added { $count } passwords. That's a great start!
+           You've added { $count } entries. That's a great start!
 }
 
 homepage-under-50-passwords =
   Hey again! Just your pal Lockie checking in!
   
-  You've added { $count } passwords. Great job, keep it up!
+  You've added { $count } entries. Great job, keep it up!
 
 homepage-over-50-passwords =
   Welcome back! I hope you're having a great day!
   
-  You've added { $count } passwords. Wow, I'm really impressed!
+  You've added { $count } entries. Wow, I'm really impressed!
 
 item-details-title = Entry Name
 item-details-origin = Website Address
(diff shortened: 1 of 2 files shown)


devinreams:
@changecourse @sandysage Let's look at getting final strings in for Alpha. You'll see each file has some that we can swap out with little-to-no-work.

I took a first pass based on what's in Invision... (click ["Files Changed"](https://github.com/mozilla-lockbox/lockbox-extension/pull/163/files) here to see)

src/webextension/locales/en-US/manage.ftl
```diff
@@ -1,22 +1,22 @@
...
-item-summary-no-username = (no username)
+item-summary-new-item = New Entry
+item-summary-no-title = (No Entry Name)
+item-summary-no-username = (No Username)
 
 homepage-no-passwords =
   Welcome to Lockbox! I'm Lockie, and I'm here to help you lock
   up your passwords!
-  
+
```
jimporter: I'm pretty sure this will break things. The trailing whitespace is important for ensuring that fluent.js parses the entire body as one string.

src/webextension/locales/en-US/manage.ftl
```diff
@@ -1,22 +1,22 @@
...
+  *[other] { $count } Entries
 }
 
-toolbar-add-item = Add Item
+toolbar-add-item = New Entry
 toolbar-go-home = Home
 toolbar-send-feedback = Feedback
 
-item-filter.placeholder = Filter…
+item-filter.placeholder = Search for an Entry
```
jimporter: I believe the prevailing style in Firefox is sentence case, not title case.

src/webextension/locales/en-US/manage.ftl
```diff
@@ -1,22 +1,22 @@
 toolbar-item-count = { $count ->
-   [one]   1 item
-  *[other] { $count } items
+   [one]   1 Entry
+  *[other] { $count } Entries
```
jimporter: I think this should also be sentence case, since it's not really a title (though this one is less clear, since button labels are title case in Firefox).

src/webextension/locales/en-US/manage.ftl
```diff
@@ -1,22 +1,22 @@
...
 
-item-filter.placeholder = Filter…
+item-filter.placeholder = Search for an Entry
 
-item-summary-new-item = (new item)
-item-summary-no-title = (no title)
-item-summary-no-username = (no username)
+item-summary-new-item = New Entry
+item-summary-no-title = (No Entry Name)
+item-summary-no-username = (No Username)
```
jimporter: Not sure about title case here... Maybe we really just need some CSS to style these specially, like as italic or something?

src/webextension/locales/en-US/manage.ftl
linuxwolf: Atom was overly helpful.  Added back in next commit.

src/webextension/locales/en-US/manage.ftl
changecourse: I was basing the decision for title case off of the photon design guidelines, here: http://design.firefox.com/photon/copy/capitalization.html

src/webextension/locales/en-US/manage.ftl
linuxwolf: I've dealt with too much internationalization concerns elsewhere to feel confident with CSS directives here (e.g., Turkish dotless i).  Maybe I shouldn't be so concerned, but I'd rather start with the strings matching the mockups directly.

jimporter: That's fine; I was just thinking about greying them out a bit or something to indicate that they're not human-entered.

linuxwolf: The italics or color change seems like something we can follow up on ... this text matches the invision mockup + how I interpret the Photon guidelines.

linuxwolf:
@jimporter @changecourse latest commit should address your concerns ... would you confirm?

devinreams:
Thank you all for working together and getting this over the finish line!

PR #164, opened 2017-10-11
Restrict the editor view to a single instance

### LINKED ISSUE #151: Only allow one editor view of lockbox open at a time

linuxwolf:
Resolves #151 

Every click on the toolbar button (once initialized and unlocked) opens the same Editor tab.  This can be easily backed out when/if it's deemed.


m8ttyB:
Functionality works, r+ from QA.

jimporter:
@linuxwolf Thoughts on outright removing `MultipleView` for the time being? We can always grab it from git history if we need it back.

linuxwolf:
@jimporter 

I can go either way; this PR went with expediency.  I can also remove it and its tests.

Issue #129, opened 2017-10-04
Update dependencies before alpha release


jimporter:
Most notable is React 16, which has a ~30% reduction in file size. That should help us load our extension's pages a bit faster.

---

- [ ] prune unpopular/unsupported packages
- [ ] review outdated packages (and decide to update them)
  - [ ] decide on updating React 16 now vs later
- [ ] review dependencies vs devdependencies for appropriateness (css-loader, style-loader)
- [ ] decide to pin lockbox-datastore to a tag?

jimporter:
Now would probably also be a good time to prune packages that aren't especially "popular", for lack of a better word. If we use more-mainstream packages, we have a greater degree of certainty that other people are finding issues with them so we're not the ones who have to stumble on random bugs. High up on the list of things to remove:

* `async-wait-until`
* `nodezip`
* `xml-webpack-plugin`
* `react-addons-test-utils` (it's deprecated!)

pdehaan:
Current list of deprecated packages:

```sh
$ npm outdated

Package                         Current      Wanted  Latest  Location
copy-webpack-plugin               4.0.1       4.1.0   4.1.0  lockbox
enzyme                            2.9.1       2.9.1   3.1.0  lockbox
eslint                            4.7.1       4.8.0   4.8.0  lockbox
eslint-plugin-node                5.1.1       5.2.0   5.2.0  lockbox
eslint-plugin-react               7.3.0       7.4.0   7.4.0  lockbox
extract-text-webpack-plugin       3.0.0       3.0.1   3.0.1  lockbox
jsdom                            11.2.0      11.3.0  11.3.0  lockbox
lockbox-datastore                 0.0.1         git     git  lockbox
mocha                             3.5.3       3.5.3   4.0.0  lockbox
mocha-webpack                1.0.0-rc.1  1.0.0-rc.1   0.7.0  lockbox
prop-types                      15.5.10      15.6.0  15.6.0  lockbox
react                            15.6.1      15.6.2  16.0.0  lockbox
react-addons-test-utils          15.6.0      15.6.2  15.6.2  lockbox
react-dom                        15.6.1      15.6.2  16.0.0  lockbox
sinon                             3.3.0       3.3.0   4.0.1  lockbox
sinon-chai                       2.13.0      2.14.0  2.14.0  lockbox
source-map-support               0.4.18      0.4.18   0.5.0  lockbox
style-loader                     0.18.2      0.18.2  0.19.0  lockbox
[...]

PR #167, opened 2017-10-11
Update dependencies

jimporter:
This updates all our dependencies and cleans a few small things up. I also added a small fix to something I did wrong in the UX branch; now the manage UI takes up the whole height of the window.


devinreams:
Fixed #129

PR #168, opened 2017-10-12
(WIP) Alpha UI updates to get closer to mockups

### LINKED ISSUE #138: Update user interfaces to match designs

### LINKED ISSUE #179: Logging in from the toolbar - Show/Hide of password field resizes window
toolbar_login

### LINKED ISSUE #190: A lot of text breaks the entry forms
screenshot 2017-10-17 10 13 34

**Steps to reproduce**
1. put a significant amount of text in each of the fields (eg lots of Lorem Ipsum)

Per glob's finding [via Slack](https://mozilla.slack.com/archives/C4LNYNJ8G/p1508206366000055).

devinreams:
Fixes #138 
Fixes #179 (side-effect of shared code)
Fixes #190 

Quick start at a branch to get closer to @changecourse mockups:

https://mozilla.invisionapp.com/share/5JDP20NYB#/screens/256782990

screen shot 2017-10-11 at 8 24 21 pm

Where it's at right now:

screen shot 2017-10-11 at 8 21 38 pm

Prioritized things to do before Alpha based on discussion with @changecourse:

- [x] update the firstrun boxes to borrow a little bit of FxA styles
- [x] stack label/input on views 
  - started but not great, needs better spacing
- [ ] add view/edit titles?
- [ ] change secondary buttons to links (save vs delete)
  - started but needs more spacing. not sure if Text is the right item to use
- [x] update the header bar weight/style to get more breathing room and visibility
- [x] placeholder for empty list

It's not done but I wanted to push this up for others to see, review, and help with as needed.

Note: the text strings (item vs entry) are addressed at #163

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,5 +32,6 @@ module.exports = {
     "comma-dangle": ["error", "always-multiline"],
     "curly": "error",
     "no-console": "warn",
+    "semi": "error",
   },
-}
+};
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -46,5 +46,5 @@ module.exports = function(config) {
     autoWatch: true,
     singleRun: true,
     concurrency: Infinity,
-  })
+  });
 };
--- a/src/webextension/firstrun/components/master-password-setup.js
+++ b/src/webextension/firstrun/components/master-password-setup.js
@@ -94,7 +94,7 @@ export default class MasterPasswordSetup extends React.Component {
         </label>
         <div className={styles.error}>{error}</div>
         <Localized id="master-password-setup-submit">
-          <Button className="default" type="submit">iNIt</Button>
+          <Button theme="primary" type="submit">iNIt</Button>
         </Localized>
       </form>
     );
--- a/src/webextension/firstrun/components/master-password-setup.js
+++ b/src/webextension/firstrun/components/master-password-setup.js
@@ -51,7 +51,7 @@ export default class MasterPasswordSetup extends React.Component {
         await browser.runtime.sendMessage({
           type: "close_view",
           name: "firstrun",
-        })
+        });
       } catch (err) {
         console.log(`initialize failed: ${err.message}`);
         // TODO: Localize this!
--- a/src/webextension/firstrun/components/master-password-setup.js
+++ b/src/webextension/firstrun/components/master-password-setup.js
@@ -17,6 +17,10 @@ export default class MasterPasswordSetup extends React.Component {
     this.state = {};
   }
 
+  componentDidMount() {
+    this._firstField.focus();
+  }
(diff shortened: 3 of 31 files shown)


jimporter:
> [buttons] started but needs more spacing. not sure if Text is the right item to use

`Text` isn't the right thing to use here; it'll likely cause issues for accessibility due to having the wrong ARIA roles. I think it'd be better to use `Button` but provide some sort of prop that tells the component how it should be themed. For instance:

```
// The bright blue primary button
<Button theme="primary"/>

// A button that looks more like a link
<Button theme="minimal"/>

// A "regular" button
<Button/>
```

m8ttyB:
When hiding/showing the password, the input fields resize.

field_resize

jimporter:
> When hiding/showing the password, the input fields resize.

Works on my machine. Presumably the macOS browser styles are doing something funky... I don't know that there's much I could do though since I don't have a Mac to test on.

m8ttyB:
How can I help you get traction on this, issue #179 outlines a similar behavior.

jimporter:
> How can I help you get traction on this, issue #179 outlines a similar behavior.

Write a patch or find someone who can reproduce the issue who wants to write a patch? There's not much I can do about a visual issue that doesn't occur for me.

m8ttyB:
Perhaps @linuxwolf can help on this.

devinreams:
>>When hiding/showing the password, the input fields resize.

>How can I help you get traction on this, issue #179 outlines a similar behavior.

FWIW the current styles are bumping the input and button together in a strange way.

I didn't dive into too deeply when I started this but I was going to re-arrange the input and the "Show/Hide" button to match the comps anyway. I think part of that re-arrangement to closely match this will make this width problem a non-issue:

screen shot 2017-10-16 at 2 29 27 pm

jimporter:
> I didn't dive into too deeply when I started this but I was going to re-arrange the input and the "Show/Hide" button to match the comps anyway. I think part of that re-arrangement to closely match this will make this width problem a non-issue:

I'm not sure that's a good idea. FxA's style is to have the show/hide button in-line with the password field, and we should probably follow their lead. We should try to be consistent across our products so people don't have to context-switch as much. (They also make you hold your mouse button down for "show", which is weird but a separate issue.)

jimporter:
@linuxwolf I think this is sufficiently done that it could be reviewed.

linuxwolf:
This change introduces an edge-case bug.  If the filter results in no entries:

screen shot 2017-10-20 at 09 05 35

If it's quick (<1 hour) to fix, then I'd say do it here.  Otherwise, we log an issue and address it then.

PR #173, opened 2017-10-12
adding folder for images and our alpha lockie

changecourse:


changecourse:
changed to lockie_v1.

PR #175, opened 2017-10-12
Fix telemetry event re-registration in the middle of an existing Firefox session

jimporter:
As discussed on Slack. Here's a quick fix that just makes our telemetry category versioned. This will break the other telemetry PR, so I'll land this after and fix stuff up myself.

### DIFF
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -8,85 +8,102 @@ const { utils: Cu } = Components;
 
 Cu.import("resource://gre/modules/Services.jsm");
 
+// In order to allow us to register new telemetry events in the middle of a
+// Firefox session, we currently need to ensure that our events have a unique
+// category name. In order to do this, every time we update the events in any
+// way, we must also give them a unique category name. If you're updating the
+// events, please increment the version number here by 1.
+const TELEMETRY_CATEGORY = "lockboxv0";
+
 function startup({webExtension}, reason) {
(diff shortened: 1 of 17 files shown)


jimporter:
I *think* the test failures are because I changed the message passing API to no longer include the `"lockbox"` category. I'm also talking with the telemetry team to see if there's a better way to do all this.

jimporter:
@linuxwolf this should be ready for re-review.

src/bootstrap.js
```diff
@@ -8,85 +8,102 @@ const { utils: Cu } = Components;
...
+        objects: ["manage"],
+        extra_keys: ["fxauid"],
+      },
+    });
+  } catch (e) {
+    if (e.message === "Attempt to register event that is already registered.") {
+      // eslint-disable-next-line no-console
+      console.log("telemetry events already registered; skipping registration");
+    } else {
+      throw e;
```
linuxwolf: I wonder if we should stop the extension from running if there are other telemetry problems.

The lack of telemetry will make @irrationalagent sad, but I don't think it's worth breaking users.  Definitely log *why*, but not sure it should stop the extension from running ...

jimporter: If there's another problem, it's (almost) certainly programmer error. The most common one would be if we used an invalid string for one of our events.

PR #186, opened 2017-10-17
Switch to using karma for tests

### LINKED ISSUE #196: Run Functional Tests in a Browser Environment

jimporter:
Fixes #196 

WIP, coverage is still broken


linuxwolf:
@jimporter do you think this is ready for review?

jimporter:
@linuxwolf Well... I'm not *happy* with this PR yet, mainly due to the problems with coverage. I have one or two ideas that might help, but if they don't work out, then I'll probably put what I have now up for review.

jimporter:
Ok, I give up. This is probably the best I can make it without spending several more days messing with istanbul.

test/mocks/xpcom.js
```diff
@@ -0,0 +1,16 @@
...
+window.Components = {
+  utils: {
+    "import": function() {},
+  },
+};
+
+window.Services = {
+  telemetry: {
+    registerEvents() {},
+    recordEvent() {},
```
linuxwolf: (for the future, since this kind of comment should also have been made long ago elsewhere)

this can be a good place to keep a recording log, and  verify the calls that are expected match what was captured.

jimporter: I don't think I'm doing it right now, but generally I just leave the functions in our mocks empty (unless we need some functionality for a test), and then use Sinon to watch for when they're called in the relevant test so I can make sure they were called with the right args.

jimporter:
Force-merged after a rebase.

PR #220, opened 2017-10-20
Use ~ for fluent dependencies

stasm:
Since fluent is still in its 0.x days, I suggest to only allow automatic
upgrades to the most recent PATCH versions. That's what the ~ operator
does in package.json. MINOR versions (allowed by ^) might introduce API
incompatibilities.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,10 +49,10 @@
     ]
   },
   "dependencies": {
-    "fluent": "^0.4.1",
-    "fluent-intl-polyfill": "^0.1.0",
-    "fluent-langneg": "^0.1.0",
-    "fluent-react": "^0.4.1",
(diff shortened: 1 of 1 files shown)


linuxwolf:
This change is not necessary.

Per [semver's section on **Caret Ranges**](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004), this **won't** update to the next minor version if the major version is 0.  For example, using our current dependency on `fluent`:

* **^0.4.1 := >=0.4.1 <0.5.0**

This means `npm update fluent` will change it from `0.4.1` to `0.4.5` but not to `0.5.0`.

stasm:
Oh, wonderful! I had no idea. This makes so much sense, thanks for pointing it out!

PR #227, opened 2017-10-20
Improve first-run tests slightly

jimporter:
This fixes a minor issue where `<MasterPasswordSetup/>` wasn't always using controlled fields in React and adds a test to get us to that coveted 100% coverage. I also fixed an issue with clearing event listeners in the bootstrap.js tests.


jimporter:
Just for future reference, I noticed that this fixes an actual bug: when setting up your master password, if you type something in one field and delete it, you won't be able to proceed because the fields don't match. That's because one field's value is `""` and the other's is `undefined`.

m8ttyB:
Great catch, functionality verified in Nightly.

PR #229, opened 2017-10-23
Ensure en-US is the default locale for dev builds; resolves #140

### LINKED ISSUE #140: Ensure en-US is the default locale for dev builds
Right now, our automatic locale listing code from `DirListWebpackPlugin` will emit our locales in (probably) alphabetical order, which is wrong; `en-US` should be first so that it gets set as the default locale. The easiest way to do this would be to just hardcode `en-US` as the default, or we could be a little fancier and allow sorting the array from `DirListWebpackPlugin`.

This won't matter until we have other locales though, so it's not urgent.

jimporter:
Resolves #140

### DIFF
--- a/dir-list-webpack-plugin.js
+++ b/dir-list-webpack-plugin.js
@@ -19,6 +19,7 @@ export default class DirListWebpackPlugin {
       directory: undefined,
       filename: undefined,
       filter: undefined,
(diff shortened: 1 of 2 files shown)


jimporter:
Landed after rebasing and checking tests.

PR #231, opened 2017-10-23
Add maxLength to <EditItemDetails/>; resolves #103

### LINKED ISSUE #103: Add front end validation of fields
Each input field should have reasonable and appropriate limits to what, and how much, data can be placed there.

Exactly where the validation happens is less important than taking steps to ensure that the user can be aware of where the limit is hit, and possibly what that limit is.

\[ for original description, see [this comment](https://github.com/mozilla-lockbox/lockbox-extension/issues/103#issuecomment-338262084)]

jimporter:
resolves #103

### DIFF
--- a/src/webextension/manage/components/edit-item-details.js
+++ b/src/webextension/manage/components/edit-item-details.js
@@ -64,9 +64,10 @@ export default class EditItemDetails extends React.Component {
 
   render() {
     const {newItem, onSave, onCancel} = this.props;
-    const controlledProps = (name) => {
+    const controlledProps = (name, maxLength = 500) => {
       return {name, value: this.state[name],
-              onChange: (e) => this.handleChange(e)};
+              onChange: (e) => this.handleChange(e),
+              maxLength: maxLength.toString()};
     };
 
     return (
@@ -107,7 +108,7 @@ export default class EditItemDetails extends React.Component {
           <Localized id="item-details-notes">
             <LabelText>nOTEs</LabelText>
           </Localized>
-          <TextArea {...controlledProps("notes")}/>
+          <TextArea {...controlledProps("notes", 10000)}/>
         </label>
         <Toolbar className={styles.buttons}>
           <Localized id={`item-details-save-${newItem ? "new" : "existing"}`}>


src/webextension/manage/components/edit-item-details.js
```diff
@@ -64,9 +64,10 @@ export default class EditItemDetails extends React.Component {
 
   render() {
     const {newItem, onSave, onCancel} = this.props;
-    const controlledProps = (name) => {
+    const controlledProps = (name, maxLength = 500) => {
```
linuxwolf: are there tests that can be done to verify the limits are enforced?

Also, I'm not sure silently truncating the input is expected by the user.  Most other apps/sites that have limits tell you when they are exceeded.

jimporter:
@linuxwolf I can add tests, but it seems a bit over the top to test that HTML attributes are getting applied by React and that Firefox respects `maxlength`.

linuxwolf:
@jimporter it seems reasonable to me to check that, after simulating the entry of too much data, what the front-end submits to the back-end meets the required limits, I think it would help detect accidental regressions.

jimporter:
@linuxwolf Ok, maybe I can't add tests... at least not *useful* ones. Since we're using Enzyme to test our React stuff, and Enzyme mocks a bunch of stuff, about the best I could come up with is to test that the `maxlength` attribute is set on the real DOM element. Setting a long string via `simulateTyping` doesn't result in truncation or even an invalid form field. If you have any ideas of how we could get the tests to do what we want though, I'm all ears.

This is something I can easily imagine an *integration* test for, though. That has the benefit of running in a really-real browser environment.

Also, in the long term, I agree we'd want to allow the user to enter more than the allowed number of characters and just alert the user that they need to get rid of some text. That's quite a bit more complex than setting an HTML attribute though. (I don't think it's *hard* exactly, but we'd probably want some UX design, and there's an opportunity cost to working on stuff like this at the moment).

linuxwolf:
> Also, in the long term, I agree we'd want to allow the user to enter more than the allowed number of characters and just alert the user that they need to get rid of some text. That's quite a bit more complex than setting an HTML attribute though. (I don't think it's hard exactly, but we'd probably want some UX design, and there's an opportunity cost to working on stuff like this at the moment).

This would be a great time to engage with @changecourse on that, but I'll accept the limitations of the testing environment for now, but really want an update to the latest datastore before 0.1.1.

PR #232, opened 2017-10-23
Disable the Firefox built-in password manager on install; resolves #184

### LINKED ISSUE #184: Adding An Item Triggers FireFox Native Password Manager
Please delete this issue if we feel that it is not a problem.

When submitting a new credential on nightly mac os, the native firefox login manager asks if you want to save what you just entered in Lockbox. See screenshot below.

screen shot 2017-10-16 at 2 28 57 pm

The only reasons that I can think of that this may be a (minor) issue:

1. The user already has a similar (conflicting?) credential in the Fx native password manager and clicking save changes it / dupes the existing entry.

2. Testers accidentally pollute their native Fx password manager with foobar / test credentials 

Both 1 and 2 of course require the user to actively allow the native PW manager to save the lockbox entry, which I feel is unlikely.

jimporter:
Note that this only applies to *new* installs of Lockbox, so existing users will want to uninstall Lockbox, restart Firefox, and reinstall. This is slightly inconvenient, but shouldn't be a problem in the long term.

### DIFF
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -2,11 +2,16 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* global ADDON_INSTALL */
 /* eslint-disable no-unused-vars */
 
 const { utils: Cu } = Components;
 
 Cu.import("resource://gre/modules/Services.jsm");
+
+const REMEMBER_SIGNONS_PREF = "signon.rememberSignons";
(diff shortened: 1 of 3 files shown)


m8ttyB:
A question, at point will the built-in password manager be disabled? Upon installation of our extension or only after the first run scenario has been completed?

linuxwolf:
At (fresh) installation time.

m8ttyB:
This also begs the question, if a user disables LB extension, will the built-in password manager be re-enabled?

linuxwolf:
No, it will not.  The original setting will be undated when it is uninstalled.

I believe this was discussed in the planning meeting.

PR #235, opened 2017-10-24
Abstract modals code so it can be used elsewhere

jimporter:
This abstracts all the code for our modals so that it can be dropped into other places more easily. `<ModalRoot/>` is now a widget, and the dialog boxes are easier to define thanks to `<LocalizedConfirmDialog/>`.


jimporter:
Merged after rebasing.

PR #237, opened 2017-10-24
Fix modal dialogs

### LINKED ISSUE #162: Clicking "Home" doesn't pop up "Save changes" modal when editing an item

jimporter:
Resolves #162 
only the last 4 commits are unique to this PR.

### DIFF
--- a/src/webextension/manage/actions.js
+++ b/src/webextension/manage/actions.js
@@ -21,6 +21,7 @@ export const SELECT_ITEM_COMPLETED = Symbol("SELECT_ITEM_COMPLETED");
 
 export const START_NEW_ITEM = Symbol("START_NEW_ITEM");
 export const EDIT_CURRENT_ITEM = Symbol("EDIT_CURRENT_ITEM");
+export const EDITOR_CHANGED = Symbol("EDITOR_CHANGED");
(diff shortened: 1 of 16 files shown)


m8ttyB:
Entering characters into a new item, deleting those character, and then selecting the `Home` button triggers the unsaved changes dialogue.

STR:
- select `New Entry` button
- enter characters into any input field, then delete those characters
- select the `Home` button

Expected:
The unsaved changes dialogue shouldn't be triggered.

unsaved_changes_hhome_button

jimporter:
> The unsaved changes dialogue shouldn't be triggered.

No, I think it should be triggered. Pretty much every application I've ever worked with does that. (Now, most will revert to "no changes" if you go back via undo, and we don't do that, but that's something we could look at later.)

m8ttyB:
Note, I'm only advocating this for unsaved, new items, that have empty form fields. Entering input into a new, unsaved item should not trigger the unsaved changes modal.

This would also bring the behavior inline with the user flow of selecting 'New Item` and then selecting `Home` without entering any form data.

empty_fields

/cc'ing @changecourse for clarificaiton

jimporter:
@m8ttyB I checked, and this happens on master too, so this isn't the best place to discuss this. Could you file a new issue for it?

PR #238, opened 2017-10-25
Fix line wrapping in <ItemDetails/>; resolves #236

### LINKED ISSUE #236: An item with a long password breaks the viewer layout
Appears related to the work completed in issue #190
 
untitled

jimporter:
resolves #236 

CSS is good.

### DIFF
--- a/src/webextension/manage/components/item-details.css
+++ b/src/webextension/manage/components/item-details.css
@@ -42,6 +42,7 @@
 
 .inline-button {
   display: flex;
+  overflow: hidden;
 }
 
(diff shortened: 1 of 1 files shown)


jimporter:
Yeah, I'm not entirely sure what our logic should be for long strings or if we want to do something like truncate the password dots (or just say 16 dots no matter what). I went with the way it is now just because the other ways looked weird for the common case and I didn't want to fuss with it too much.

jimporter:
Merged after rebasing (didn't wait for tests since it's a one-line CSS change).

PR #239, opened 2017-10-25
Localize page titles; resolves #225

### LINKED ISSUE #225: Localize page titles
Right now, the page titles for all our documents are unlocalized, and I'm not sure how to localize them via `fluent-react`.

@stasm What's the best way to localize the page title with `fluent-react`? The title element is "outside" of our React code, so `<Localized/>` and `withLocalization` probably won't help.

jimporter:
resolves #225 

Now our page titles are localized! vOv

The one minor downside is that we get a brief flash of the URL in the tab bar before our localizations load, but that's probably only noticeable if you're really paying attention.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11695,6 +11695,16 @@
         "prop-types": "15.6.0"
       }
     },
+    "react-document-title": {
+      "version": "2.0.3",
+      "resolved": "https://registry.npmjs.org/react-document-title/-/react-document-title-2.0.3.tgz",
+      "integrity": "sha1-u/kioNcUEvyUgkXkKDskEt9w8rk=",
+      "dev": true,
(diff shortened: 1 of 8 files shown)


linuxwolf:
Although this might be a couple of big assumptions ...

If the majority of users will have a locale that is the same as our default locale -- most users might not ever see a flash if the localization replacement is identical to the default, if a default were still set in `webpack.babel.config`.

jimporter:
Hmm... on the one hand, I'm worried that we'd let the strings get out of sync, although perhaps this wouldn't be a big deal anyway. Longer-term, we could do some stuff to make this even better; I'll file a followup about it.

jimporter:
Merged after rebasing and waiting for tests to pass.

Issue #240, opened 2017-10-25
Be careful with withLocalization()/getString()


jimporter:
As @stasm wisely pointed out in #225, using `withLocalization()`/`getString()` won't respond to changes to the user's locale. We use this in a few places, and we should either get rid of them or listen to language changes (though I'm not sure how we'll be doing that just yet...).

stasm:
The goal of `getString` was to provide an imperative way to retrieve translations for one-off use-cases: `alert()`, modal windows, push notifications etc. They are usually displayed once and dismissed after use. The translation doesn't persist in the UI and it doesn't need to react to language changes.

jimporter:
There are also a couple places where we use unlocalized text that I can resolve pretty easily as part of this.

sandysage:
I've changed this issue to more broadly account for dynamic text.

@jimporter Is there more to do as part of this story to localize for dynamic text?

PR #243, opened 2017-10-25
Remove usage of withLocalization()/getString() from the item list

jimporter:
I'd like to remove `withLocalization()`/`getString()` wherever possible. This removes it for the item list.

@stasm Is there a better way to handle the localization here? I want to have a fallback string when the user-defined string is empty, but the only way I could get to work was to pass the string's length in separately and select based on that.

### DIFF
--- a/src/webextension/locales/en-US/manage.ftl
+++ b/src/webextension/locales/en-US/manage.ftl
@@ -20,9 +20,17 @@ all-items-empty =
   Create a new entry so Lockie has something to protect.
 all-items-filtered = No results
 
-item-summary-new-item = New Entry
-item-summary-no-title = (No Entry Name)
-item-summary-no-username = (No Username)
+item-summary-new-title = New Entry
+item-summary-title =
+  { $length ->
+     [0]     (No Entry Name)
+    *[other] { $title }
+  }
+item-summary-username =
+  { $length ->
+     [0]     (No Username)
+    *[other] { username }
+  }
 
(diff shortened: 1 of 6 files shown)


devinreams:
@jimporter I marked this as "fixes" #240 as I didn't see any immediate follow-on task(s). if there are, just for our sanity and organization, can you please open a new issue(s) describing the todo(s)?

jimporter:
@devinreams This doesn't fix #240 (which is why I didn't say so in the commits; I usually try to flag the issues I'm resolving on each commit, though occasionally I forget). Let's just leave #240 open and I can post a quick summary of what's left.

jimporter:
@linuxwolf This could use another quick review, since the rebase had some significant merge conflicts and I had to change some stuff.

m8ttyB:
Is there a requirement for PI to verify this or are the checks that were done by @linuxwolf et al sufficient? If this looks correct locally to others, I'm okay with a r+.

jimporter:
@m8ttyB Since we only have one locale at the moment, there's nothing really to verify (except that things generally work). When we have more than one locale, we'll want to verify that changing the locale (assuming we support that in some way) updates all parts of the UI, including the list items' placeholder strings.

devinreams:
>Let's just leave #240 open and I can post a quick summary of what's left.

Ah, sorry I misunderstood the intention here and work remaining. I thought I was being helpful! 😔

PR #246, opened 2017-10-26
Make `nsp` a dev-dependency so it can be run locally (and so we're no…

jimporter:
…t just assuming Travis will always have nsp installed)


jimporter:
Note: one potential downside of this is that it puts us on the hook for remembering to keep `nsp` up to date, although I don't know how important it is that we always stay on the latest-greatest.

PR #247, opened 2017-10-26
Remove some unnecessary console.log calls

jimporter:
This just removes some logging and disables warnings for some other ones. I'm totally willing to debate which logging should be kept around, but I mainly went with just keeping error logging around.

With this patch, there are 0 eslint warnings, so maybe we could even make `console.log` usage throw an eslint error now. (It'll still warn on local builds, since we run lint in warn-only mode there.)


m8ttyB:
Do you need QA to test this?

jimporter:
Nah, it's just logging.

PR #248, opened 2017-10-26
Update dependencies to all the latest versions

jimporter:
It's almost time for another release, so let's update our dependencies! Tested locally and everything seems to work (and tests pass, of course).


m8ttyB:
What's the risk associated with this, what value does it add? 

Will we need to do full run of the e2e tests? If so, I'd like hold this until next release.

jimporter:
They're all minor/patch releases (except codecov, which is a major release), and all except `react-modal` are dev-only packages. `react-modal` has some small enhancements to its focusing logic that shouldn't affect us.

m8ttyB:
Today is code freeze, unless it solves a known bug or adds required functionality, I'm r-.

devinreams:
Good points made here. And as discussed today, how about we agree:

- watch for back-end updates that need to be updated as needed (sooner than later)
- do dependency updates at _beginning_ of dev cycle to catch any unexpected results/bugs

@linuxwolf to re-review/clear.

PR #249, opened 2017-10-26
Restore focus after closing modals

### LINKED ISSUE #133: Focus is lost when focused element goes away
If you're focused on something and it goes away (e.g. you close the item editor or close a modal dialog), the focus is lost entirely, so you have to hit Tab to get to anything. We should be better about this. I'm not 100% sure what the best way is though, since manually updating focus at all the relevant times seems fairly brittle.

- [x] for modals: #249 
- [ ] for changing edit/view mode

jimporter:
This partially resolves #133. Maybe it's worth adding a test, but on some level it would just be testing that `react-modal` works right (also, it might make more sense for such a test to be in our not-yet-existent integration tests).


jimporter:
Clearing review until I make the tests happy.

jimporter:
Ok, tests should pass. There aren't any tests for the focus restoration because we can't use `react-modal` in our tests yet. We have to stub it because `enzyme` doesn't yet support React 16 portals; hence, there's currently no way to test that we're using `react-modal`'s focus logic correctly.

devinreams:
Need latest to fix tests.

Then ready for review.

jimporter:
@devinreams @linuxwolf So, it turns out I pushed the version of my code with passing tests, but Travis decided to run the old code and mark the new code as having failed, despite them not having the same hashes.

I hate Travis.

jimporter:
Ohhhh, it's because we added a new place using Modals and those tests failed. I didn't realize that Github rebased PRs onto master before submitting them to Travis.

PR #251, opened 2017-10-27
Use rjsx-mode for JS files in src/webextension

jimporter:
This makes my life easier (and presumably the lives of other emacs users). It should only trigger if you have js2-mode installed, but assumes you also have rjsx-mode installed.


m8ttyB:
Given that this is a specific to dev workflows, correct me if I'm wrong, I don't believe this needs PI verification.

jimporter:
Not unless PI wants to use emacs to view our source files. :)

m8ttyB:
@jimporter I'm vim user ;-)

vim

PR #275, opened 2017-10-31
Add a warning to first run introduction

### LINKED ISSUE #270: Update Alpha first-run message
Per our discussion in Slack, we need to update the first-run page of Alpha to be:

You've just installed the Lockbox browser extension! This Alpha prototype will give you the ability to create new entries, and then later you can view, search, edit and delete those entries.

_This is a prototype that is rapidly evolving with many great improvements. Much of the functionality seen in this Alpha is expected to change. As such, data stored here may not be retained in future updates._

Please be sure to let us know your thoughts using our feedback button within the tool, including any issues you may find, things you like, and the things you're looking forward to in the future.

- [x] update the strings to include a new one "welcome-warning"
- [x] include emphasis on the welcome first run template

devinreams:
Fixes #270 requested by @sandysage to explain features may change and data may not be retained between releases.

Here's how it looks (bold section in middle is new):

screen shot 2017-10-31 at 12 24 07 pm

### DIFF
--- a/src/webextension/firstrun/components/welcome.js
+++ b/src/webextension/firstrun/components/welcome.js
@@ -20,6 +20,13 @@ export default function Welcome() {
         adipiscing elit. Etiam porta sem malesuada magna mollis euismod. Cras
         mattis consectetur purus sit amet fermentum.</p>
       </Localized>
+      <Localized id="welcome-warning">
+        <p style={{fontWeight: "bold"}}>Lorem ipsum dolor sit amet, consectetur.
+        Mauris, aliquam vel pellentesque et, mattis bibendum tellus. Fusce
+        sodales, tellus a auctor accumsan, diam risus pharetra orci, at lacinia
+        libero eros ut erat. Fusce ex neque, pharetra id rhoncus in,
+        pellentesque quis urna.</p>
+      </Localized>
       <Localized id="welcome-feedback">
         <p>Curabitur blandit tempus porttitor. Nulla vitae elit libero, a
         pharetra augue. Vestibulum id ligula porta felis euismod semper.
--- a/src/webextension/firstrun/components/welcome.js
+++ b/src/webextension/firstrun/components/welcome.js
@@ -20,13 +20,15 @@ export default function Welcome() {
         adipiscing elit. Etiam porta sem malesuada magna mollis euismod. Cras
         mattis consectetur purus sit amet fermentum.</p>
       </Localized>
-      <Localized id="welcome-warning">
-        <p><strong>Lorem ipsum dolor sit amet, consectetur.
-        Mauris, aliquam vel pellentesque et, mattis bibendum tellus. Fusce
-        sodales, tellus a auctor accumsan, diam risus pharetra orci, at lacinia
-        libero eros ut erat. Fusce ex neque, pharetra id rhoncus in,
-        pellentesque quis urna.</strong></p>
-      </Localized>
+      <p>
+        <Localized id="welcome-warning">
+          <strong>Lorem ipsum dolor sit amet, consectetur.
+          Mauris, aliquam vel pellentesque et, mattis bibendum tellus. Fusce
+          sodales, tellus a auctor accumsan, diam risus pharetra orci, at lacinia
+          libero eros ut erat. Fusce ex neque, pharetra id rhoncus in,
+          pellentesque quis urna.</strong>
+        </Localized>
+      </p>
       <Localized id="welcome-feedback">
         <p>Curabitur blandit tempus porttitor. Nulla vitae elit libero, a
         pharetra augue. Vestibulum id ligula porta felis euismod semper.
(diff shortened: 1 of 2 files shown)


devinreams:
@sandysage Should we consider updating the README and/or any other docs at this time to make sure the expectations/warnings are adequately expressed? 😁

sandysage:
I'll plan to update:
* [release notes](https://github.com/mozilla-lockbox/lockbox-extension/blob/master/docs/release-notes.md)
* [read-me](https://github.com/mozilla-lockbox/lockbox-extension/blob/master/README.md)
* [index.md](https://github.com/mozilla-lockbox/lockbox-extension/blob/master/docs/index.md)

devinreams:
@jimporter @linuxwolf are you two OK with me merging this as-is?

@sandysage would like us to then push a point release to production (with just this change) ASAP so anyone who downloads the extension has a good chance of seeing this notice (in addition to the places in docs that will also get updated separately).

Discussed in Slack: https://mozilla.slack.com/archives/C4LNYNJ8G/p1509403229000106

src/webextension/firstrun/components/welcome.js
```diff
@@ -20,6 +20,13 @@ export default function Welcome() {
         adipiscing elit. Etiam porta sem malesuada magna mollis euismod. Cras
         mattis consectetur purus sit amet fermentum.</p>
       </Localized>
+      <Localized id="welcome-warning">
+        <p style={{fontWeight: "bold"}}>Lorem ipsum dolor sit amet, consectetur.
```
jimporter: Let's not use `style`. We should do something like `<p><strong>` here instead. That should also help with blind users, since they might(?) hear something about `<strong>` but almost certainly won't hear anything about `style`.

src/webextension/locales/en-US/firstrun.ftl
```diff
@@ -9,10 +9,15 @@ welcome-intro =
     new entries, and then later you can view, search, edit, and
     delete those entries.
 
+welcome-warning =
+    This is a prototype that is rapidly evolving with many great improvements.
+    Much of the functionality seen in this Alpha is expected to change. As such,
+    data stored here may not be retained in future updates.
```
jimporter: I sort of feel like this is burying the lede... while it's nice that it's in bold (so people are likely to read this first), the actual *important* bit of this text is the third sentence, and is a little bit indirect. If we're not guaranteeing future retention of users' data (and we think people would expect us to), then I think we should be more direct about it. Something like the following would make me feel more comfortable:

> Don't rely solely on this extension for storing your passwords! This is a prototype which is rapidly evolving, and as such we can't guarantee that data will be retained in future updates.

linuxwolf: I think this suggested language is too scary.  If there's concern the current text is not explicit enough, my suggestion would be:

> This is a prototype that is rapidly evolving with many great improvements.
> Much of the functionality seen in this Alpha is expected to change. As such,
> data stored here is not guaranteed to be retained in future updates.

If there's concern the text is also possibly too verbose, might I suggest this:

> This is a rapidly evolving prototype that will change radically as Alpha progresses.
> Any data stored here is not guaranteed to be retained in future updates.

devinreams: Good input, thanks, you two.

Ultimately this is @sandysage's call, especially as she's been consulting with other trusted folks on this as well...

sandysage: Given these points, let's trim the message to be:

> This is a rapidly evolving prototype that will change as Alpha progresses. Any data stored here is not guaranteed to be retained in future updates.

Which only states the warning: doesn't bury the lede, but also doesn't over-emphasize the risk involved.

src/webextension/firstrun/components/welcome.js
devinreams: @jimporter Agreed on use of strong.

But after a few minutes of searching, I'm not sure how to make [Localized](https://github.com/projectfluent/fluent.js/wiki/Localized) take both the `p` and `strong` tags. It drops the `strong` once localized.

See my latest commits versus actual output when run:

screen shot 2017-11-01 at 10 32 35 am

src/webextension/firstrun/components/welcome.js
```diff
@@ -20,6 +20,13 @@ export default function Welcome() {
         adipiscing elit. Etiam porta sem malesuada magna mollis euismod. Cras
         mattis consectetur purus sit amet fermentum.</p>
       </Localized>
+      <Localized id="welcome-warning">
+        <p><strong>Lorem ipsum dolor sit amet, consectetur.
```
linuxwolf: The fastest fix to to keep `<strong>` is to move `<Localized>` inside the <p>`, so that `<strong>` is an immediate child.

devinreams: Yeaaaaah, I was hoping there was a simple elegant solution. I'll drop that in for now and wait until later today to see if @jimporter has a different preference.

screen shot 2017-11-01 at 10 59 42 am


Pretty low risk here since this will likely all change and it's more important just to push this out and get it in front of any new users ASAP.

m8ttyB:
r+ from PI. Verified consistent on OSX and Win10, Firefox Nightly.

PR #278, opened 2017-11-01
Update editor tab URL when toolbar clicked

### LINKED ISSUE #262: Clicking Lockbox icon does not open editor, if editor was previously opened and the URL changed
1. open Lockbox editor
2. enter a new URL in toolbar (eg reddit.com)
3. navigate away, leaving that tab/window open that was the lockbox editor but was changed to a different URL
4. click the Lockbox icon
5. focus changes back to tab/window that was originally the editor, showing the new site and does not open a new editor
lockbox-editor-does-not-open

linuxwolf:
Fixes #262

This fix changes the URL for the tab assigned to the Editor (or
First-Time) is to be displayed.  This change does not:

* mess with the tab's history (`back`/`forward` buttons)
* keep whatever the user was originally navigated to


m8ttyB:
I think this experience still needs a bit of work. If a user uses the Lockbox tab to browse to a page, clicking the toolbar icon should open a new tab, not reuse the tab it previously had focus in.

STR:
0. open lockbox
1. use the tab lockbox is open in to browse to a new URL
3. click the lockbox toolbar icon

Expected:
A new tab containing lockbox is opened, leaving the content in the previous tab untouched.

Actual:
Lockbox is opened in the tab containing the content the user was viewing.
back_button

linuxwolf:
@m8ttyB 

I went back and forth and around.  My feelings are increasingly moving toward "we shouldn't be able to navigate away or into Lockbox content".  I am uncomfortable with the ability of users to go into their history and open the Lockbox Entries (or Lockbox First-run), or navigate away and potentially back.

This PR is essentially maintaining that status quo.

m8ttyB:
>  My feelings are .. "we shouldn't be able to navigate away or into Lockbox content"

r+ agreed

jimporter:
> "we shouldn't be able to navigate away or into Lockbox content"

It's not clear to me why this is important for the management UI (it matters for first-run, but we can do something else for that UI). While I can see some potential for there being caching issues (e.g. user navigates away from the management page, locks the store, and then hits back and gets a cached page with the store unlocked), we can test that to make sure it's not happening. However, I do think it's important that we don't navigate away from arbitrary pages the user has opened, since we can't know what they were doing at the time, and we might lose data (e.g. filled forms) as a result. Even though the user could hit "back" and (usually) get their form data back, I doubt everyone is going to realize that.

Unfortunately, I can't r+ a patch that adds the potential for dataloss (or even just the feeling of dataloss) on other sites. The only options I can see to move forward on this are 1) open a new tab and do what we need to feel secure about the management UI being in tab history, or 2) open a chromeless window for the management UI.

src/webextension/background/views.js
```diff
@@ -9,19 +9,22 @@ export class SingletonView {
   }
 
   async open() {
-    let windowId;
+    let windowId,
+        offPath = false,
```
jimporter: I don't have a strong opinion, but maybe call this `navigatedAway`?

linuxwolf:
I think the longer-term approach can be addressed as part of #285 or its successor.

PR #284, opened 2017-11-02
Always request review from PI

devinreams:
After discussion with @m8ttyB I'd like to propose we default to _always_ automatically request review from the @mozilla-lockbox/product-integrity team on all code on all PRs.

This way we can be explicit about sign-offs per feature before they go into the `master` branch (which we intend to be stage-able or deploy-able at any time). That way any manual test steps are added/accounted for and performed and unit/integration tests are considered.

I open it up to you all, what do you think? I'd like to incorporate a decision or any changes discussed here into my process doc draft before we meet as a team on Monday.

### DIFF
--- a/docs/CODEOWNERS
+++ b/docs/CODEOWNERS
@@ -1,9 +1,6 @@
 # lockbox-extension code owners
 # https://help.github.com/articles/about-codeowners/
 
-# tests should also be reviewed by @m8ttyB
-/test/    @jimporter @m8ttyB
-
 # docs are shepherded by @devinreams
 /docs/    @jimporter @devinreams
 
--- a/docs/CODEOWNERS
+++ b/docs/CODEOWNERS
@@ -1,9 +1,6 @@
 # lockbox-extension code owners
 # https://help.github.com/articles/about-codeowners/
 
-# docs are shepherded by @devinreams
-/docs/    @jimporter @devinreams
-
 # include Flod for l10n string changes
 # @todo uncomment below after initial string freeze
 #src/webextension/locales/en-US/*.ftl    @flodolo
--- a/docs/CODEOWNERS
+++ b/docs/CODEOWNERS
@@ -11,5 +11,6 @@
 # @todo uncomment below after initial string freeze
 #src/webextension/locales/en-US/*.ftl    @flodolo
 
-# otherwise, everything is to be reviewed by @jimporter
-*         @jimporter
+# everything is to be reviewed by @jimporter
+# also, default to always include PI
+*         @jimporter @mozilla-lockbox/product-integrity


jimporter:
I don't think (pre-landing) review is the right phase for QA. While I support the idea of master always being in a releasable state, I also think that adding more review hurdles for landing patches runs the risk of slowing us down too much. Ideally, we'd have QA look at patches as they're about to land, so that we have verification of all the changes made during code review. However, that adds extra time for every patch before it can land, and there are already occasions where patches pile up. If we go too far with allowing the pileup, we increase the risk of merge conflicts, which can result in another review cycle to verify that the conflict was resolved correctly.

How about a process more like we use on Bugzilla? In Bugzilla, patches that land result in the bug being set to RESOLVED, and then when QA looks at it, they update the state to VERIFIED. While that's not quite as pure as having master always be releasable, it still provides an easy status indicator for if we've done QA (just check for merged-but-not-verified PRs).

m8ttyB:
A question for our Github gurus, what tooling exists that exposes a flow similar to Buzilla's status states?

devinreams:
>In Bugzilla, patches that land result in the bug being set to RESOLVED, and then when QA looks at it, they update the state to VERIFIED

Good call out. I was thinking about this "after the fact" approach to confirming, too and forgot to mention it in my thinking above.

@m8ttyB to your question on the specifics, since issues have binary (open/closed) states the best option is to use labels consistently.

Two proposals:

- the _presence_ of a label (`needs-pi`) then gives you an easy population of things to review and is explicit and consistent with how we may flag things needing input from others. One problem being making sure they always get applied and enforcing that.

- The opposite of course is the _absence_ of a label and applying it after verified/confirmed. The population to review then _every_ PR (which is what this is proposing anyway) and afterwards apply a `pi-verified` (or similar) label. Which could then be excluded from searches ("show me all closed PRs that don't have the verified label").

m8ttyB:
The "after the fact" approach is a well worn trail that can add value and ends in a simple to execute workflow. My concerns are several. Merging work into the master branch injects risk if that feature isn't adequately tested near code freeze on Thursday. The team needs enough time to integrate feedback from PI on features before we can ship them.

tl;dr, my proposal. PI tracks work labeled for the milestone that the team is working on. No additional labels are required (`needs-pi`). The dev team can merge feature branches that have met their standards (and pass the automated checks). Post-merge PI will verify the feature. Once a feature has been verified by PI and the test suite(s) have been updated to reflect the new functionality, PI will set a `pi-verified` (or similar) label. 

The caveat(s):
* The team strives to do sprint planning and earmark work with the milestone it is expected to land in.
* A triaged backlog is maintained
* If immediate or special attention is needed on a feature by PI, the team will use the `pi-needed` (or similar) label. I don't want the notion of after-the-fact testing to become the standard we rely on. Offering timely, collaborative feedback is important when a product is rapidly being developed.
* `pi-needed` should be used for PRs that require PI review prior to merging. Additionally you can add me to a review.
* PI will continue to track open PRs and will continue to engage on code reviews.

devinreams:
>Merging work into the master branch injects risk if that feature isn't adequately tested near code freeze on Thursday. The team needs enough time to integrate feedback from PI on features before we can ship them.

Agreed, as we've all acknowledged, but maybe worth stating again: the downside of not preventing a merge is the risk the branch is not as stable as we'd like.

I _believe_ we as a group have agreed we more heavily weight the ability to "move fast and break things" (ugh, overused, I know 😔) in the early stages of the project and agree with your proposal:

>Post-merge PI will verify the feature. Once a feature has been verified by PI and the test suite(s) have been updated to reflect the new functionality, PI will set a pi-verified (or similar) label.

👍 

The caveats are also understood and agreeable, as we touched on in our team meeting yesterday. 👍 

@mozilla-lockbox/engineering any final comments on any of the above ☝️ before I close this and capture the decisions in our process doc?

devinreams:
This "after the fact" approach will be documented at https://github.com/mozilla-lockbox/mozilla-lockbox.github.io/issues/10 instead of a code owner/required upfront review in GitHub as originally proposed here. 🙅 

Thanks for the discussion, ya'll. 🎩

Issue #255, opened 2017-10-30
Photon Typography


changecourse:
In lieu of extensions.css not providing us the Photon style for free, we can begin implementing our our visual design to align with Photon.

One such area is Typography, and we can begin by adding Fira: https://mozilla.github.io/Fira/ 

My thought is that we can do one foundational effort around the implementation of the fonts, and then a subsequent pass on different weights for specific areas. 

Fonts used thus far: ExtraLight, Light, Regular, Medium, Bold

changecourse:
We can do an additional issue in the future to highlight key component definitions for typography on a 'per component' basis.

jimporter:
The spec says we use different fonts on different platforms. Shouldn't we follow that? http://design.firefox.com/photon/visuals/typography.html

Fira is for "web" stuff, and while we're all HTML, our add-on isn't really a part of the World Wide Web...

jimporter:
We *should* probably use Fira Sans Mono though, since it's an actually-good monospace font that distinguishes similar glyphs. This will make life easier on people when they need to manually type in a password.

changecourse:
Following up on this... and after some back n' forth with some of the designers tied to the Photon effort, it looks like you are right @jimporter, we should use system-based fonts (Segoe, SF, etc) for the extension, as identified in the brand guidelines list above. Looks like Fira will be more tied to other efforts around marketing, as well as the monospace need you had mentioned for passwords.

changecourse:
@jimporter that being said, we may also want to default our monospace to the platform defaults (Consolas and Menlo) instead of Fira's monospace for performance reasons perhaps?

jimporter:
@changecourse I'd strongly recommend against using system defaults for the monospace font. Many monospace fonts just aren't very good for our purposes, and I think we want to provide some guarantees that our monospace font is as readable as possible. In particular, Windows defaults to Consolas, which has precious little different between `1` and `l` (that's number one and lower-case L). Since users may occasionally have to manually type in passwords, we should make sure that it's easy to distinguish each character.

changecourse:
good to know, thanks!

changecourse:
Latest need as I see it:

- [ ] Use of SF Text on Mac instances of Lockbox
- [ ] Use of Segoe UI on Windows instances of Lockbox
- [ ] Use of Ubuntu on Linux instances of Lockbox
- [ ] User of Fira monospace for passwords throughout all instances (?)

[...]

PR #295, opened 2017-11-06
Convert to Photon theme

jimporter:
I decided it'd make more sense to just do the Photon conversion in one PR. Here it is.

I also snuck in a commit to use rjsx-mode in emacs for the `test/` directory. I forgot about that originally.


src/webextension/manifest.json.tpl
```diff
@@ -16,7 +16,6 @@
   },
 
   "browser_action": {
-    "browser_style": true,
```
devinreams: Interestingly, we get warnings and it seems that setting this to "false" (instead of removing it entirely) is expected/preferred?

Output from `jpm run`:

>1510092213233	addons.xpi	WARN	Please specify whether you want browser_style or not in your options_ui options.
>1510092213877	addons.webextension.lockbox@mozilla.com	WARN	Please specify whether you want browser_style or not in your browser_action options.

devinreams:
For posterity, here's what the updated interfaces look like on Mac:

## First Run

screen shot 2017-11-07 at 3 05 50 pm

## Home / Landing

screen shot 2017-11-07 at 3 06 30 pm

## New/Edit Entry

screen shot 2017-11-07 at 3 07 16 pm

## View Entry

screen shot 2017-11-07 at 3 07 46 pm

jimporter:
> Input fields need to be 30px high, currently 28px.

They are 30px. There's a 1px border. (We're using context-box sizing.)

> Padding between field label and field needs to be increased on welcome screen (but could be addressed with the typography story instead)

Yeah, I plan to do this in another patch, since we don't have a "label" widget yet.

> Text selection color, currently orange, should it be blue? (http://design.firefox.com/photon/components/input-fields.html#behaviors)

I don't see orange; I see the default highlight color from the OS (green in my case). Looking at the browser UI, I think that's actually what we want, although it does mean that the blue border for focused form fields is wrong (it should pull from the OS too).

> Form fields in Create a New entry are indented and shouldn't be (margin-inline-start: 7px;)

As with the other label stuff, that's for a different patch.

> Show/Hide field vs. SVG iconography

Again, for a different patch (if at all). FxA already uses a "show" string, and I think it'd be nice if we could remain consistent.

> Need definitions for puffy class of both primary and secondary btn states (New Entry and Home on latest designs)

I don't know what this means. Although I do notice that I set the font-sizes for puffy and micro wrong.

> Do we want the cursor state over a button to be different than pointer?

The current behavior follows what all the other buttons in Firefox do.

> Update Cancel to be secondary button style

Will do.

linuxwolf:
@jimporter  @changecourse The unlock doorhanger looks a little too squished.  I realize this isn't sticking around for much longer, so is it ok like this?

image

devinreams:
@jimporter since I'm already pushing pixels around lately, please let me know if you'd like me to pick up and help knock out any of the little nits here (or anywhere, for that matter).

devinreams:
@jimporter whoops, I accidentally clicked the "Update branch" button here and merged in master. sorry about that..

jimporter:
@devinreams I can probably just force-push to restore things.

jimporter:
Ok, this should be done. Last commit has all the changes.

devinreams:
Tested locally and looks good here. ✅ 

@changecourse I looked back at your comments and believe they've been a) addressed or b) deferred to the typography #255 work. is that right?

## Unlock popup

screen shot 2017-11-09 at 2 27 34 pm

## Field input measurements

screen shot 2017-11-09 at 2 32 03 pm

(28px, 30px total with border)

## Button states

screen shot 2017-11-09 at 2 33 53 pm

screen shot 2017-11-09 at 2 34 12 pm

jimporter:
@devinreams @changecourse I added two minor changes: 1) the password/search fields fonts were too small. I also made the "copy" buttons for username/password "micro" sized so that they don't mess up the relative heights of the rows.

devinreams:
>1) the password/search fields fonts were too small. I also made the "copy" buttons for username/password "micro" sized so that they don't mess up the relative heights of the rows.

@jimporter I see that in the second-to-last commit: ee6545d842d24e74e2ad3c63b6b4bbe1b8f0ec10

... but the last commit appears to undo that change? e2f086af57bf77d0cad5b1a7bd5102deebb1b77e

src/webextension/firstrun/components/app.css
```diff
@@ -5,7 +5,10 @@
 body {
   padding: 0;
   margin: 0;
+  color: #0c0c0d;
   background-color: #f9f9fa;
+  font: caption;
+  -moz-user-select: none;
```
devinreams: Mentioned at https://github.com/mozilla-lockbox/lockbox-extension/issues/311 this prevents users from selecting text such as the web address/domain on the entry view.

pdehaan:
Possibly unrelated to everything, but I notice in the [screenshot above](https://github.com/mozilla-lockbox/lockbox-extension/pull/295#issuecomment-343299179) it says "Enter your password" in the doorhanger. Shouldn't that say "Enter your master password" to be consistent with the firstrun experience?

das-first-run

vs:

jimporter:
@devinreams I made the copy buttons normal-sized again because the Photon input spec changed and now input boxes are the same height as normal-sized buttons.

devinreams:
>I made the copy buttons normal-sized again because the Photon input spec changed and now input boxes are the same height as normal-sized buttons.

Ahhhh I see now. Cool. 

@changecourse the code still looks good to me and the result appears as expected. Did you want to look again and (hopefully) approve since @jimporter pushed some small tweaks?

jimporter:
@pdehaan Can you file a separate issue for that?

PR #298, opened 2017-11-08
Replace PNG of Lockie with transparent SVG

### LINKED ISSUE #191: SVG/PNG approach for images (Lockie)
I've provided three different assets (png, png@2x, and svg) for the Lockie image in mozilla-lockbox/assets to see what the best approach will be for images long term, per a conversation with @linuxwolf 

Please let me know if you require anything else. 

At the very least, I'd love for us to replace our current png on the welcome screen with this new, transparent png, to resolve the background color issue.

devinreams:
Fixes #191 

Here's the old PNG (with a background) on the left and new hotness (transparent SVG) on the right:

screen shot 2017-11-08 at 12 16 52 pm

Note the grey clouds on the right (grey on grey background).

But also note the nice hi-DPI treatment and lack of jagged edges (looks great on my MacBook Pro).


linuxwolf:
I think the SVG still needs some fixes.  The lack of aliasing is nice, but it's not obvious those squiggles on the right are clouds.

changecourse:
After speaking with Sean Martell, we agreed that he should create an alternate version of the SVG for use on lighter colored backgrounds (which will resolve the clouds) 

As soon as I get the new asset, we can replace accordingly.

linuxwolf:
@changecourse is it worth replacing the old/busted png with the not-quite-right svg now, or wait for the ideal svg?

changecourse:
The work done by @devinreams supports SVG, so let's go with that for now, then replace SVGs once I get the new asset.

devinreams:
The SVG gets us transparency and more pixels so I'd make the argument this is an overall improvement. Easy to swap out again later.

@changecourse 👍 ❓

PR #300, opened 2017-11-08
Remove Lockie from welcome screen

### LINKED ISSUE #296: Update Lockie
I think it's time to do another pass to update Lockie. In particular, I'm thinking about the following things:

1. Remove the ASCII art
2. Remove Lockie's name from the UI (it's a goofy pun and I've always considered it more of an Easter Egg than an "official" thing)

@changecourse Thoughts?

devinreams:
Fixes #296 

This quick PR:

- removes the Lockie ASCII art (for now? ...)
- removes the speech bubble
- changes the copy to not reference Lockie

Here's the **before** state:

screen shot 2017-11-08 at 1 45 58 pm

Here's what **[recent designs](https://app.zeplin.io/project/59fb4868c5e7608593f76965/screen/5a00b68c8fa5ed3ea9452add)** show (but also assume functionality that doesn't exist, not final copy):

screen shot 2017-11-08 at 1 30 52 pm

So that's why **what you see in this PR** is just a page trim this down to it's essence, copy mostly as-was and styles as-was, knowing we still need to come back to it shortly:

screen shot 2017-11-08 at 1 43 55 pm

What else should we do/change with the goal of getting this into the next point release (freeze is tomorrow).


src/webextension/locales/en-US/manage.ftl
```diff
@@ -16,8 +16,8 @@ item-filter
 
 all-items-empty =
   Looks like you don't have any entries saved yet…
-  
-  Create a new entry so Lockie has something to protect.
+
```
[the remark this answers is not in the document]
devinreams: Good catch, thanks.

And thanks, Atom.🖕 Just changed my settings so I don't do this again...

devinreams:
>There are some l10n issues
>I'm not sure if we want to keep the strings for various password counts either...

Good catch. Perhaps we do just strip those out now, anyway and make this super simple...

It's sad to remove _some_ of this feature and whimsy but probably for the best based on what we hear from UR/UR anyway.

>and you'll also want to remove raw-loader from package.json and our Webpack config

Good call. And it's not used anywhere else, right? Didn't see anything so I removed it from our dev dependencies too. Here comes a big package-lock.json commit. 😔 

>It'd also be nice to add the big title text.

Fair enough, I re-worked this PR to drop that in. BUT...

1. I can probably do better than doing both _Localized_ and _withLocalization_ for the conditional, just wasn't sure how
2. I'm not confident in a good way to **bolden** the entry number so left it alone for now
3. I left just a shared "message" to appear beneath the title since we'll have to work in a way to show if FxA is connected or not. Consider this placeholder and _intentionally_ appearing on all screens, not just the welcome.

Here's what we have now:

screen shot 2017-11-08 at 3 03 32 pm

pdehaan:
Not sure if we want to run the images/SVGs through a minifier before committing, or if we want some build script to optimize assets.

For example, running the lockie_v2.svg file through [**svgo**](http://npm.im/svgo) reduced the image 56% (from ~76KB to 33KB):

```sh
> $(npm bin)/svgo lockie_v2.svg

lockie_v2.svg:
Done in 255 ms!
75.981 KiB - 56.3% = 33.199 KiB
```

src/webextension/manage/components/homepage.js
```diff
@@ -3,49 +3,32 @@
...
-    key = "homepage-over-50-passwords";
+    key = "homepage-has-passwords-title";
   }
 
   return (
-    <pre className={styles.homepage}>{
-      speechBubble(getString(key, {count})) + "\n" + lockie
-    }</pre>
+    <article className={styles.homepage}>
+      <img src={imgSrc}/>
```
pdehaan: nit: Do we want to add an `alt=""` tag here, or are we doing an a11y pass later?

jimporter:
I don't think it's worth thinking about minification right now. We'll probably get that for free when we start using Webpack to reference images.

src/webextension/manage/components/homepage.js
devinreams: I'm not sure what we'd say for this image specifically. We kinda intentionally don't want to "spoil" the visual joke that we're using a Loch-ness (lock) monster. 😬

devinreams:
>>Not sure if we want to run the images/SVGs through a minifier before committing, or if we want some build script to optimize assets.

Good point @pdehaan.

>I don't think it's worth thinking about minification right now. We'll probably get that for free when we start using Webpack to reference images.

@jimporter Agreed, I opened an issue here so we don't forget: https://github.com/mozilla-lockbox/lockbox-extension/issues/301

Do you spot anything else in my branch here that should be addressed? @changecourse gave the UI 👍 but I ended up dismissing it with another commit...

src/webextension/manage/components/homepage.js
jimporter: Let's do as @pdehaan suggests. `alt=""` (literally using the empty string, not filling it in like `alt="The Lockie Empire will reign for 1,000 years"`) is the recommended way of indicating that an image is purely-decoratinve.

devinreams: >See: https://www.w3.org/WAI/tutorials/images/decorative/

Ah! Sorry I misunderstood. TIL...

giphy

src/webextension/locales/en-US/manage.ftl
```diff
@@ -32,32 +31,23 @@ item-summary-username =
     *[other] { username }
   }
 
-homepage-no-passwords =
-  Welcome to Lockbox! I'm Lockie, and I'm here to help you lock
-  up your passwords!
+homepage-no-passwords-title = Welcome to Lockbox
```
jimporter: This can be merged with `homepage-has-passwords-title` below.

src/webextension/manage/components/homepage.js
```diff
@@ -3,49 +3,32 @@
...
   }
 
   return (
-    <pre className={styles.homepage}>{
-      speechBubble(getString(key, {count})) + "\n" + lockie
-    }</pre>
+    <article className={styles.homepage}>
+      <img src={imgSrc} alt=""/>
+        <h1>{
+          getString(key, {count})
```
jimporter: Let's get rid of `getString()` and use `<Localized>` instead. I should have done that earlier, but the speech bubble code was all temporary, and `getString()` was the simplest way to do it at the time...

devinreams:
>Looks good, except we can simplify the localization for the homepage title.

>Let's get rid of getString() and use <Localized> instead.

@jimporter Cool, good points. All cleaned up now.

How's that look? Ready to rock? 🚢

src/webextension/manage/components/app.css
```diff
@@ -18,8 +18,8 @@ body > main,
 
 .side-toolbar,
 .main-toolbar {
-  background-color: #f9f9f9;
-  border-bottom: 1px solid #d6d6d6;
+  background-color: #ededf0;
+  border-bottom: solid 1px #c8c8c8;
```
jimporter: Micro-nit: Let's put the border-width first.

src/webextension/manage/components/homepage.css
```diff
@@ -3,6 +3,28 @@
...
+  width: 701px;
+}
+
+.homepage > h1,
+.homepage > p {
+  color: #0c0c0d;
+  letter-spacing: 0.2px;
+  line-height: 1.5;  
+}
+.homepage h1 {
```
jimporter: Nit: Add a newline before this line.

src/webextension/manage/components/homepage.css
```diff
@@ -3,6 +3,28 @@
...
+  font-size: 33px;
+  font-weight: 300;
+  margin: auto;
+  text-align: center;
+}
+.homepage p {
```
jimporter: Nit: Add a newline before this line.

PR #302, opened 2017-11-09
Add links to overall Lockbox product/project website and docs

devinreams:
Two goals:

1. No matter how someone learns about Lockbox, I'd like to add links up to the overall product website so folks can see the whole picture (as its not all here): https://mozilla-lockbox.github.io/

2. It's also worth linking to the Waffle board so folks can see the overall project plan, priorities, etc. This small badge on the readme seemed like an OK place to add that link.

I'm open to other suggestions on how to achieve the above. Here's what it looks like in the README.

lockbox-extension_readme_md_at_10b386abd2250ac88838313eddfda369433f8518_ _mozilla-lockbox_lockbox-extension_ _github_-__private_browsing_

I also added the website blurb to the top of `docs/index.md` so its on our project website.

Once approved and merged I'll do the same (similar) to the datastore and ios repos.


devinreams:
I'm going to merge this as-is to help clean up some open PRs. I'd rather have the pointers there now and we can always go back to make tweaks later. 🚀

PR #307, opened 2017-11-09
Use Photon fonts

jimporter:
Last two commits are the actual change.


devinreams:
@jimporter This is looking good to me. Moving the labels to widgets was a good call. 👍 

@changecourse can you take a look and let us know any other tweaks you'd expect before we merge this in?

screen shot 2017-11-10 at 6 23 54 am

screen shot 2017-11-10 at 6 23 38 am

src/webextension/firstrun/components/app.css
```diff
@@ -5,7 +5,11 @@
 body {
   padding: 0;
   margin: 0;
+  color: #0c0c0d;
   background-color: #f9f9fa;
+  font: caption;
+  font-size: 13px;
+  -moz-user-select: none;
```
pdehaan: Should we pre-emptively add the non-prefixed version as well?

devinreams: I'm not entirely convinced we event want `user-select` due to its unexpected side-effects (unable to highlight/select content from entries). What do you think @jimporter?

Whatever you all decide I still think this is OK to ship as-is since we are just running it in Firefox for now and it's still marked "experimental". Good eye, though @pdehaan.

jimporter: We do, but we want `<FieldText/>` to enable selection. Since the spec is still a draft though, I'm not sure how much value there is in adding the non-prefixed version. There's really no guarantee it works like the moz-prefixed one, and it's already changed in some ways from how browsers implement it.

PR #323, opened 2017-11-11
Add pointer cursors to buttons so they look more clickable

pdehaan:
This should make all generic buttons and show/hide toggle buttons have a pointer cursor to make them look invitingly clickable.


changecourse:
Agreed with @jimporter, we had discussed a couple weeks back this very issue, but elected to align ourselves with the rest of Firefox, including current Test Pilot experiments, to not switch from standard to pointer.

PR #335, opened 2017-11-13
Bump Firefox minimum versions to 57

### LINKED ISSUE #326: A few rogue references to Firefox 54 in the repo
Per https://github.com/mozilla-lockbox/lockbox-extension/search?utf8=%E2%9C%93&q=54&type=

But I thought our min version was 57...

https://github.com/mozilla-lockbox/lockbox-extension/blob/f68fd65458599b46bdf1bb4897ccc9d2aa1dd7f5/.babelrc#L4

https://github.com/mozilla-lockbox/lockbox-extension/blob/3d5c28488fb96f348049abb16f13c6718d28f7ea/src/install.rdf.ejs#L21

https://github.com/mozilla-lockbox/lockbox-extension/blob/3d5c28488fb96f348049abb16f13c6718d28f7ea/package.json#L10

devinreams:
Fixes #326 

Note: I don't believe this will actually work until _tomorrow_ since 57 is not yet a version available to babelrc, from what I can tell. In other words, when I set the `targets: browsers` var to 57 clicking the browser action does nothing. When I set it to 56 it works. 😁


.babelrc
```diff
@@ -1,7 +1,7 @@
 {
   "presets": [
     "react",
-    ["env", {"targets": {"browsers": ["firefox >= 54"]}}]
```
devinreams: @jimporter this doesn't work presumably because nothing is JS expected to be "babel"-ed. should we just keep this at 54? looking at [plugins.json](https://github.com/babel/babel-preset-env/blob/master/data/plugins.json) maybe consider 55?

Issue #293, opened 2017-11-06
First-run create new entry


sandysage:
No-navigation presented with first entry: https://mozilla.invisionapp.com/share/MHEBCD4PK#/screens/262155635

sandysage:
Closing as duplicate of #292

PR #337, opened 2017-11-14
Add landing page create button

henryhahn:
This PR adds a button to the landing page for being able to directly add a new entry.

This is meant to start implementing the designs associated with #293.

screenshot 2017-11-14 01 10 20

### DIFF
--- a/src/webextension/locales/en-US/firstrun.ftl
+++ b/src/webextension/locales/en-US/firstrun.ftl
@@ -18,6 +18,8 @@ welcome-feedback =
     button within the tool, including any issues you may find, things
     you like, and the things you're looking forward to in the future.
 
+welcome-add-item = Create Entry
+
 master-password-setup-formtitle = Please create a master password below:
 
 master-password-setup-password = Password
--- a/src/webextension/locales/en-US/firstrun.ftl
+++ b/src/webextension/locales/en-US/firstrun.ftl
@@ -18,8 +18,6 @@ welcome-feedback =
     button within the tool, including any issues you may find, things
     you like, and the things you're looking forward to in the future.
 
-welcome-add-item = Create Entry
-
 master-password-setup-formtitle = Please create a master password below:
 
 master-password-setup-password = Password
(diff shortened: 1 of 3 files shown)


src/webextension/locales/en-US/firstrun.ftl
```diff
@@ -18,6 +18,8 @@ welcome-feedback =
     button within the tool, including any issues you may find, things
     you like, and the things you're looking forward to in the future.
 
+welcome-add-item = Create Entry
```
pdehaan: Is this string used somewhere?

henryhahn: Oops, good catch.

henryhahn:
Hey @jimporter - I'm looking at the mocks linked in the bug: https://mozilla.invisionapp.com/share/MHEBCD4PK#/screens/262155636

It does look to me like its meant to happen here below the fold. Do you mind clarifying what you mean?

src/webextension/locales/en-US/firstrun.ftl
henryhahn: Oops! Removed.

devinreams:
Thanks for the contributions thus far, @henryhahn!

Right now @changecourse (our UX expert and designer) is re-imaginging some of the flows from "first run" through to "coming back to a welcome screen" (aka landing page) and what those should be to guide the user's experience. Long story short: we have a lot of moving parts and the designs aren't all ready _quite_ yet.

We expect some explicit requirements (in the form of a new issue) to come soon to help clearly describe what we want to change. Stay tuned!

jimporter:
@henryhahn The mockup you linked shows different buttons on the homepage than the one you added. Based on the mockup, we plan to put buttons there to set up a Firefox Account for sync/master password stuff. The button to create a new entry stays at the top left as it is now.

devinreams:
@henryhahn Thanks again for your contributions. I admit we were a bit premature in pushing up concepts that weren't fully baked for implementation (and not obvious online here that was the case).

Going forward we'll explicitly tag "good first issue" to match the GitHub convention. As the project matures we'll have more and more. But for now, we've taken a different user experience approach and this PR no longer applies.

Issue #331, opened 2017-11-13
Photon Updates, 2nd Pass


changecourse:
~Depends on #338 (styles moving around a lot)~

Work at https://github.com/mozilla-lockbox/lockbox-extension/pull/351

Follow-up at #392 

Now that the design has firmed up a bit more, there are additional tasks to be completed for alignment to the prototype:

- [x] Navigation links (top right) to match the photon guidelines, found [here](http://design.firefox.com/photon/components/link.html) including external link icon
- [x] Vertical height of Nav bar, 72px
- [x] Moving entries counter into primary content area (You have __ entries in your Lockbox)
- [x] Moving home next to "New Entry" and applying secondary button styling to it
- [x] Sidebar width (320px)
- [x] Styling of new entry in sidebar to match with content background of main content area, based on screenshot below: <img width="583" alt="screen shot 2017-11-13 at 11 43 35 am" src="https://user-images.githubusercontent.com/6164801/32742898-ff9c4cee-c867-11e7-9e80-7bbce505f46b.png">
- [x] Removal of horizontal divider in Create a New Entry & View Entry Details<img width="385" alt="screen shot 2017-11-13 at 11 44 51 am" src="https://user-images.githubusercontent.com/6164801/32742940-20ba234c-c868-11e7-9223-3c0e8793316a.png">
- [x] Change background color of primary content area when viewing or creating an entry to hex F3F3F5, as seen in latest design comps. 
- [x] Fix password and search input font color so its legible (very light, reported #264)
  <img width="681" alt="screen shot 2017-11-14 at 10 09 03 am" src="https://user-images.githubusercontent.com/49511/32793670-de8343b6-c923-11e7-9f55-b02f404e6f03.png">

---

Welcome and Homepage

- ~Disabled button states for both primary (blue) and secondary (grey)~ #361
- [x] 1.1.0 welcome screen width and fonts
- [ ] 1.1.0 zero entries / first run home page with 3-up feature heros
- ~Disabled "Home" button when not on a item detail (create, edit, view) view~ #361 

New Entry and Sidebar

- ~1.1.0 New Entry: Disable buttons for New Entry while creating a current entry (may be accounted for somewhere else)~ #361
- [x] revisit button sizes
- [x] "Site Address" and "Site Name" order and copy
- ~Disable "Save Entry" until at least one field has been populated (may be accounted for somewhere else)~ #361 
[...]

Issue #206, opened 2017-10-19
View entries in a doorhanger


sandysage:
When accessing an online account, I need to be able to reliably log in with my saved credentials. If that does not autofill the form, then I need to be able to copy, and even see, the password to manually enter those details into the website form.

# Acceptance Criteria
1. Viewing an entry details takes 3 clicks or less
2. Viewing all the entries or a single entry doesn't take the user out of focus from the web form
3. Read view of entries is unencumbered by any other action, including "create a new entry"

# Expected behavior
1. Pre-req: assumes you have a Lockbox account created and at least 1 entry saved
2. Navigate to Lockbox by clicking the browser action
3. Default view has full list of entries, by Entry Name, ordered alphabetically
4. Click the entry name to see the entry details
5. Entry details has option to "show" or "copy"
6. Clicking out of the doorhanger, closes it

# Visual Design
https://mozilla.invisionapp.com/share/MHEBCD4PK#/262323598_D_1-1-0_Default_-_Entries_List 

# Context (research, metrics, finding)
Accessing accounts is when efficiency is most required. Research indicates that users might change behavior (consider or take a different action than the default), but not when accessing accounts, sensitive or not.

Prior issue was created here: https://github.com/mozilla-lockbox/lockbox-extension/issues/112

# Relevant user stories

- [x] Implement initial doorhanger code #338
- [x] Add Telemetry events to doorhanger #358

sandysage:
- Prereq: the database is unlocked/signed-in
- Click on the toolbar icon
- List of entries appears in a doorhanger
- Click an entry to see details
- Click edit to open the main editor, in a tab, to that entry in edit mode
- Click add new to open the main editor, in a tab, to a blank new entry
- Click fullscreen to go to the editor, in whatever state it's current at
- Search behaves the same as in the fullscreen editor, just in the doorhanger
- Click Sign out locks the database

![screenshot-2017-10-27 2 1 0 lockbox home popover - returning user - entries list](https://user-images.githubusercontent.com/850418/32126876-3dd6c5e2-bb30-11e7-952f-5b175290572c.png)

changecourse:
@jimporter I just added the visual design to the description above. Would love to get your review/input ahead of implementation. Also designing additional doorhanger states over the next couple of days.

jimporter:
[...]

PR #338, opened 2017-11-14
Add a doorhanger

jimporter:
- [x] doorhanger
- [x] re-arrange the webextension files
- [x] redo some components for better styling
- [x] tests
- [x] move l10n around
- [x] go to doorhanger from unlock prompt
- [x] pref to disable at build-time

Starts #206

Blocks #331


devinreams:
That. Is. Awesome.

Awesome

For those playing along at home, here's what it looks like:

- [x] click browser action to open list view
- [x] search items from list view
- [x] click/select item to enter detail view
- [x] copy username/password from detail view

doorhanger

@jimporter What else do you have on your list to get this on master for us to start testing and tweaking and building more against?

jrbenny35:
This PR is kinda huge.

m8ttyB:
@jimporter this looks amazing. Nice work! Looking forward to the refactored work once it's out of WIP.

devinreams:
Quick notes from today's status update:

- the WIP doorhanger commit is ready for review (definitely a big change, follow on work is planned)
- the grand re-arrangement / refactor commit is there (looks like HUGE change, really just moving lines)
- to the first point, we'll get this wrapped in a pref/option so we can merge the progress into master and then continue on smaller branches/PRs from there
- some unit test coverage needed before ready for final merge but @jimporter will tag folks since this the above is ready to review in-progress work

devinreams:
>100% of diff hit

Well done, @jimporter! :heart:

jimporter:
Ok, this isn't quite ready to land, but it should be ready for some preliminary reviews.

src/webextension/list/index.js
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
linuxwolf: It feels odd to have an `index.js` that's not an export of public things in `/list`.

src/webextension/list/common.js
```diff
@@ -1,7 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
linuxwolf: This is a "six of one, half dozen of the other" kind of thing, but I would have expected `flattenItem` / `unflattenItem` to move here.

src/webextension/manage/actions.js
```diff
@@ -87,11 +87,12 @@ function addItemStarting(actionId, item) {
   };
 }
 
-function addItemCompleted(actionId, item) {
+function addItemCompleted(actionId, item, interactive = false) {
   return {
     type: ADD_ITEM_COMPLETED,
     actionId,
     item,
+    interactive,
```
linuxwolf: It took me a bit to see what `interactive` meant, but I get it now.  I can't think of a better term, though.  It might help to have a comment somewhere with the intent, in case the "us" in six months struggle.

devinreams:
>this is done

Nice. Very nice. Just did a quick re-run and confirmed...

- [x] `npm run run` _does_ load with doorhanger
- [x] `npm run package` (production build) does _not_ show the doorhanger
- [x] all screens are localized (firstrun, welcome, item list, item details) in both doorhanger and full

Looks great from a functional perspective. And from our quick team discussion today we are quite happy with how this has turned out. I know it required a lot of hard work "behind the scenes", thanks again @jimporter. 🥇

jimporter:
@m8ttyB Just a small note: coverage will be slightly reduced until we remove the pre-doorhanger code. Technically, we could run a second test job to get that coverage back in the short term, but I'm not sure if it's worth it. If you think we should, I can file an issue and whip up a patch to land after the integration tests.

m8ttyB:
@jimporter Nah, I think we're good. The risk rightfully low given this is pref'd off. Let's save for a later date, prior to the public debut.

jimporter:
@m8ttyB Just to be 100% clear, right now we're *not* testing the code path that will ship to users. We *are* testing the doorhanger code path, since that's the future, though.

PR #342, opened 2017-11-14
Update Metrics.md and add events

irrationalagent:
I've updated metrics.md to accurately reflect what we are capturing, removed outdated information, and tidied things up. I've also included an outline of doorhanger-related events for when we get there. 

I've also added three new events, `item_copied` for when a credential is copied to the clipboard, `resetRequested` triggered when someone clicks to reset the datastore thru settings, and `resetConfirmed` when the datastore reset is actually confirmed. 

I played around with a filter box event, but since the event fired on every keypress I thought it would be better not to include it for now. If anyone has any idea for a once-only filter box event then feel free to chime in.


docs/metrics.md
```diff
@@ -163,294 +117,51 @@ These are the metrics we plan to collect regarding the state of user datastores.
...
-                    "item_id": "new"});
-```
+1. `fxaSignIn` fires when someone clicks the signin button during firstrun. **objects**: fxaSignInPage.
 
-### Events Related to Changes in the Datastore
-We will record an event of **category** `lockbox_datastore.changed` when there is a change to the datastore. These events will be recorded as `lockbox_entry_changed`.
+2. `confirmPW` fires when the user clicks the button to confirm their FxA pw. **objects**: confirmPWButton
 
-**Methods** for this event are `added`, `updated` and `deleted`.
+3. `doorhangerItemSelected` fires when a user clicks an item in the doorhanger's itemlist. **objects** doorhanger
```
devinreams: @jimporter in case you haven't seen it yet, here are the events that @irrationalagent has defined for the doorhanger #338

src/webextension/manage/components/item-details.js
```diff
@@ -14,9 +14,12 @@ import Toolbar from "../../widgets/toolbar";
 
 import styles from "./item-details.css";
 
+import * as telemetry from "../../telemetry";
+
 const PASSWORD_DOT = "\u25cf";
 
 function CopyToClipboardButton({text, ...props}) {
+  telemetry.recordEvent("itemCopied", "entryDetails");
```
devinreams: This is going to fire the "itemCopied" event every time the button is drawn on the page, not when actually clicked.

src/webextension/settings/actions.js
```diff
@@ -31,13 +33,15 @@ function localResetStarting(actionId) {
 }
 
 function localResetCompleted(actionId) {
+  telemetry.recordEvent("resetCompleted", "settings");
```
devinreams: The reset request and completion were captured as expected for me 👍 

telemetry_data_and_update_metrics_md_and_add_events_by_irrationalagent_ _pull_request__342_ _mozilla-lockbox_lockbox-extension

devinreams:
@irrationalagent can you check out my latest commit https://github.com/mozilla-lockbox/lockbox-extension/pull/342/commits/12fd4b8d3bdf17810284d6029d322559281031eb and see if that fires and records all the data you're expecting now?

Once you've confirmed its looking good locally I think we just need a final 👍 from @jimporter on both the metrics.md update and the events themselves. 😀

irrationalagent:
hmm good catch @devinreams. probably didnt catch it the first time because i always copied when i tested. TIL to make sure to test by not doing things.

your changes work for me locally

docs/metrics.md
```diff
@@ -144,14 +96,16 @@ See the Events section for specific examples of event registration and recording
...
 
 -   The setup flow, so we can know at what points (if any) people quit the flow before finishing it
--   Top-level interactions centered around use of the Lockbox toolbar icon. This includes interactions within the initial pop-over that is displayed when the user clicks the icon.
--   Interactions with the main CRUD editor view which lists the user's Lockbox items (credentials)
+-   Top-level interactions centered around use of the Lockbox toolbar icon. This includes interactions within the initial doorhanger that is displayed when the user clicks the icon.
+-   Interactions with the list of the user's Lockbox items (credentials)
 -   Interactions with the add / modify dialogs used to enter / edit item information
 -   Changes to the datastore that actually contains the user's items, in addition to user actions that lead to those changes
 -   When the user submits feedback about lockbox
+-		Usage of the copy and reveal functions for stored Lockbox items.
```
pdehaan: Indenting is wacky here, causing this to be converted into a code block. 🤷‍♀️

docs/metrics.md
```diff
@@ -163,294 +117,51 @@ These are the metrics we plan to collect regarding the state of user datastores.
...
-For events related to the submission of actual item information, see the next section. The item editors have 5 fields: `title_field`, `origin_field` (URL for the credential), `username_field`, `password_field`, and `notes_field`. There are also three buttons: `save_entry_button`, `cancel_button`, and `toggle_password_button`. The latter toggles the visibility of characters in the password field.  
+-   `timestamp_last` The timestamp of the last edit the user made to the datastore. Does not necessarily correspond to the last time they opened the CRUD editor.
 
+-   `n_uses` The number of times a user copied or auto-filled a Lockbox item.
 
-The **methods** are `click` and `render`.
+## List of Events Currently Recorded
 
-Example event registration (Substitute `lockbox_new_item` with `lockbox_edit_item` to register events for the existing item editor):
+All events are currently implemented under the **category: lockboxV0**. The `extra` field always contains `fxaid` where possible (i.e. after FxA auth). For events pertaining to a particular Lockbox item, `itemid` is also included. They are listed and grouped together based on the contents of the event's `method` field.
```
pdehaan: Should this be "fxaid" or "fxauid"?

devinreams: Good catch, I'll change this too.

docs/metrics.md
devinreams: Thanks. I'll go fix this right now @pdehaan @irrationalagent

devinreams:
So I don't forget: I discussed with @jimporter that the missing test coverage on this patch (the Telemetry event oncopy itself) is nice-to-have and in line with the current coverage for the project (does Telemetry work: yes or no).

docs/metrics.md
```diff
@@ -85,18 +37,18 @@ We will aim to collect data that will help us understand the following (please a
...
-- Maybe: what type of sites do people use Lockbox with **(not in alpha)**?
+- What type of sites do people use Lockbox with?
+    - Do they use Lockbox only for their most sensitive accounts?
 
 ## Collection
 
 **Note:** *This is the collection plan for our internal alpha release. For our beta release will be taking advantage of test pilot's telemetry API. The metrics plan for beta will be described in a separate document.*
 
-At this point, all measurements related to Lockbox will be made client-side. All users will have to authenticate through FxA, and thus additional measurements related to that will be logged on the FxA auth server. We have no control over what is already collected via that mechanism.
+At this point, all measurements related to Lockbox will be made client-side. Eventually all users will have to authenticate through FxA, and thus additional measurements related to that will be logged on the FxA auth server. We have no control over what is already collected via that mechanism.
```
jimporter: Minor point, but I thought our goal was to turn FxA into an option that we let users upgrade to later.

src/webextension/manage/components/item-details.js
```diff
@@ -14,11 +14,14 @@ import Toolbar from "../../widgets/toolbar";
...
 
+import * as telemetry from "../../telemetry";
+
 const PASSWORD_DOT = "\u25cf";
 
 function CopyToClipboardButton({text, ...props}) {
   return (
-    <CopyToClipboard text={text}>
+    <CopyToClipboard text={text}
+      onCopy={() => telemetry.recordEvent("itemCopied", "entryDetails")}>
```
jimporter: Nit: the indentation here is a bit off. I think something like the following would be better:

```js
    <CopyToClipboard text={text} onCopy={() => {
      telemetry.recordEvent("itemCopied", "entryDetails");
    }}>
```

Otherwise, the `onCopy` attribute should be lined up below `text`, but that'll probably push us past 80 columns.

docs/metrics.md
devinreams: >I thought our goal was to turn FxA into an option that we let users upgrade to later.


@jimporter Good point, it is not expected to be always required.

@irrationalagent Based on our discussion this morning, do you want to take a pass at re-wording this? I can if you're busy.

src/webextension/manage/components/item-details.js
devinreams: @jimporter How does that look now? 

https://github.com/mozilla-lockbox/lockbox-extension/blob/7f11411c44ad9fb25d707ced84186198764995ff/src/webextension/manage/components/item-details.js#L21-L29

I feel like I should indent line 25 (`}}>`) since the `Button` is nested within. but it also feels wrong. Your call. 🙃  

(and if there's a style guide I can go read, please point me to it!)

devinreams:
@jimporter Alright. This is ready for final review!

Do you want it to merge before or after the doorhanger? I'm happy to clean this up on this side if its easier that way.

As we chatted as a team earlier today, we're planning to land this (and the doorhanger, pref off in production builds) within a day so that we can push another point release and start tracking these new events. 🤘

devinreams:
Please note: due to the nature of the shared components between the full editor _and_ doorhanger (hidden, available only in dev modes) Telemetry events in the doorhanger fire identically to those in the full editor.

Follow-up issue to address those is here: https://github.com/mozilla-lockbox/lockbox-extension/issues/358

@jimporter I merged in the latest changes so this is ready for your final review, please. 😀

devinreams:
@jimporter Will you have a few moments to review this today? I'd like to get the release process started ASAP. 🚂

src/webextension/list/components/item-fields.js
```diff
@@ -16,11 +16,15 @@ import TextArea from "../../widgets/text-area";
...
 
+import * as telemetry from "../../telemetry";
+
 const PASSWORD_DOT = "\u25cf";
 
 function CopyToClipboardButton({text, ...props}) {
   return (
-    <CopyToClipboard text={text}>
+    <CopyToClipboard text={text} onCopy={() => {
+      telemetry.recordEvent("itemCopied", "entryDetails");
```
jimporter: Two questions and a comment:

1. Do we want to distinguish between copying the username and copying the password? That might give us useful info.
2. Do we want to (eventually) distinguish between doing this in the management UI and the doorhanger? `entryDetails` doesn't seem especially useful at the moment, since seeing the copy button automatically means you're looking at an entry's details.
3. Finally, if we do use `entryDetails`, let's rename it to `itemDetails`. Internally, we call these things "items", but the UI calls them "entries" (this is slightly confusing, but I chose "item" to match the datastore's API, and the user-facing strings could really be anything).

irrationalagent: 1. Yes, that would be ideal. Just couldn't quite figure the best way to implement that, so I did this as a quick first solution. If you could give me a quick pointer on where might be best to implement these as distinct events, I could do it before this is merged.

2. That's the plan for the next iteration, once I get my head around the doorhanger changes. no need to do it here I think. If we keep the "itemCopied" nomenclature for the event method, then we can add the changes to the event object in the next version to distinguish between the two and we should be fine w/r/t data continuity. 

3. I'm fine with renaming entryDetails to itemDetails for this PR. I'll do it

jimporter: For (1), easiest would be something like this:

```js
function CopyToClipboardButton({text, field, ...props}) {
  return (
    <CopyToClipboard text={text} onCopy={() => {
      telemetry.recordEvent(`${field}Copied`, "entryDetails");
    }}>
      <Button {...props}/>
    </CopyToClipboard>
  );
}

CopyToClipboardButton.propTypes = {
  text: PropTypes.string.isRequired,
  field: PropTypes.oneOf(["username", "password"]).isRequired,
};
```

jimporter: Oh, and I forgot, you'd have to use `<CopyToClipboardButton text={blah} field="username"/>`

irrationalagent:
thanks for the help @jimporter. I added the username/pw specific copy events and testing locally they seem to work.

screen shot 2017-11-30 at 3 21 31 pm

devinreams:
@irrationalagent this is now live in production, you should start seeing the new events trickle in on pings soon 📣

PR #346, opened 2017-11-16
npm update dependencies for 0.1.3-alpha

### LINKED ISSUE #343: Update and test dependencies for 0.1.3-alpha
While we make a final decision on what/how to do this #276 #304 the fact remains:

We should update and review our dependencies for the next release ASAP so everything is tested against it as early and often as possible.

devinreams:
Resolves #343 

Tests pass and this seems to run as expected on my machine still.

But, the bump to estlint resulted in a 84 `Missing space before function parentheses` warnings, so I changed [the rule to "never"](https://eslint.org/docs/2.0.0/rules/space-before-function-paren#always) which, looking quickly at our code, I believe is accurate and expected.


m8ttyB:
I did a quick run of the e2e tests on OSX, lgtm. :

.eslintrc.js
```diff
@@ -33,5 +33,6 @@ module.exports = {
     "curly": "error",
     "no-console": "warn",
     "semi": "error",
+    "space-before-function-paren": [2, "never"],
```
pdehaan: Nit: Me thinks this should be `["error", "never"]` for consistency.

I think this is coming from code w/ anonymous functions like `async()`.
Curiously, this seems to be coming from some change in **eslint-plugin-mozilla**, if I'm looking at this correctly.

jimporter: If it's all async arrow functions that it's grumpy about, we should fix the code. I only avoided putting a space there to make the linter happy in the first place.

pdehaan: In that case, running `npm run lint -- --fix` should auto fix all these to these:

```diff
-openDataStore().then(async(ds) => {
+openDataStore().then(async (ds) => {
```

devinreams:
@pdehaan @jimporter here's what `--fix` resulted in: https://github.com/mozilla-lockbox/lockbox-extension/pull/346/commits/dd04c57649b7876d526702aacaac1c9e5f8f4cc7 (all changes to asyncs)

How does that look to you all? 🙃

pdehaan:
LGTM, but I'll let @jimporter give a 👀 and push the shiny green button.

PR #359, opened 2017-11-29
Update to more precise and accurate favicon paths

devinreams:
These moved at #338 and so the favicon on the full manage page is missing.

The firstrun happens to be right, but `list/manage` is no longer one-level away from the icons.

Either way, `icons` is now top-level so we can assume that here, I believe.

### DIFF
--- a/webpack.config.babel.js
+++ b/webpack.config.babel.js
@@ -144,7 +144,7 @@ export default {
       chunks: ["webextension/firstrun/index"],
       inject: false,
       minify: htmlMinifyOptions,
-      icon: "../icons/lb_locked.svg",
+      icon: "/icons/lb_locked.svg",
     }),
     new HTMLWebpackPlugin({
       template: "template.ejs",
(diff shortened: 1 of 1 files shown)


devinreams:
>I think I'd rather leave the unlocked icon around

Pardon my confusion, do you mean "leave around" as in simply keep the asset? That we have in the `assets` repo. Or do you mean to keep the unlocked icon state showing your datastore is unlocked?

Either way we do have to balance our individual opinions with the informed decisions from the research and design folks....

jimporter:
I mean that I think we should revert that commit for the time being, until we have the functionality that that icon is supposed to tie into (i.e. the ability to re-lock the datastore). Without that functionality, I agree that the dual icons are pretty useless, but I'm not convinced we can validate it either way in our research until we have the corresponding functionality to test against.

devinreams:
>I mean that I think we should revert that commit for the time being

@jimporter Fair, it's more important to fix the bug than try to tack on "undoing" the functionality here anyway.

Unless I hear otherwise I'll merge this in tomorrow in prep for 0.1.3 release.

PR #361, opened 2017-11-30
Disable the filter when there are no items; resolves #324

### LINKED ISSUE #324: Hide or disable the entry search form if user has zero entries?
Current behavior feels a bit weird (being able to search with zero entries). Not sure if this is the latest UI spec, but it looks like maybe we should be disabling the form in the zero entries state:
https://mozilla.invisionapp.com/share/5JDP20NYB#/screens/256782985

lockbox_entries

jimporter:
Here's a quick version of disabling the filter when there are no items. No tests yet; I just wanted to see what people think of the UX, since the Photon specs don't appear to have anything about what disabled form fields should look like. We could also add a different placeholder string if we want, e.g. "Add some entries before filtering them" (that's terrible wording, but you get the idea).

- [x] Disable search filter input / clear when no items exist
- [x] CSS for disabled button states for both primary (blue) and secondary (grey) buttons
- [ ] Disabled "Home" button when not on a item detail (create, edit, view) view (aka can't click "Home" when home)
- [x] Disable button for "New Entry" while actually creating a current entry
- [ ] Disable "Save Entry" until at least one field has been populated

### DIFF
--- a/src/webextension/list/containers/item-filter.js
+++ b/src/webextension/list/containers/item-filter.js
@@ -20,6 +20,7 @@ function ItemFilter(props) {
 export default connect(
   (state) => ({
     value: state.filter,
+    disabled: state.cache.items.length === 0,
   }),
   (dispatch) => ({
     onChange: (value) => { dispatch(filterItems(value)); },
--- a/src/webextension/list/manage/containers/add-item.js
+++ b/src/webextension/list/manage/containers/add-item.js
@@ -9,17 +9,13 @@ import { connect } from "react-redux";
 
 import Button from "../../../widgets/button";
 import { startNewItem } from "../../actions";
+import { NEW_ITEM_ID } from "../../common";
 import * as telemetry from "../../../telemetry";
 
-function AddItem({dispatch}) {
-  const doClick = () => {
-    telemetry.recordEvent("addClick", "addButton");
-    dispatch(startNewItem());
-  };
-
+function AddItem({disabled, onAddItem}) {
   return (
     <Localized id="toolbar-add-item">
-      <Button theme="primary" onClick={doClick}>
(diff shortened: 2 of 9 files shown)


devinreams:
@changecourse here's what it looks like when disabled (both doorhanger and full editor):

screen shot 2017-11-30 at 3 13 33 pm

The mouse cursor also shows the disabled/not-allowed indicator:

screen shot 2017-11-30 at 3 15 56 pm

@jimporter Both the prop and CSS both look good to me and appreciate seeing this. I'd like to see if I can take this same approach with some of the Photon changes requested at #331.

devinreams:
Other disabled states that @jimporter and I discussed from https://github.com/mozilla-lockbox/lockbox-extension/issues/331 that may as well be considered / included here:

- [ ] CSS for disabled button states for both primary (blue) and secondary (grey) buttons
- [ ] Disabled "Home" button when not on a item detail (create, edit, view) view (aka can't click "Home" when home)
- [ ] Disable button for "New Entry" while actually creating a current entry
- [ ] Disable "Save Entry" until at least one field has been populated

I added them to the original description of the PR above to keep track of. If it doesn't make sense to do any of these here, we can change it.

jimporter:
> Disabled "Home" button when not on a item detail (create, edit, view) view (aka can't click "Home" when home)

This one in particular seems odd to me, since it'd be sort of like disabling the Home button when on the homepage in Firefox itself. I'm also worried that the change in color would be distracting for people. vOv

devinreams:
@m8ttyB thanks for this. I meant to move these sooner but have just added the two remaining open items to #392 to discuss and address in a future sprint.

m8ttyB:
Woot, thanks @devinreams. Bumping this to verified.

PR #368, opened 2017-12-04
update dependencies for 0.1.5-alpha

devinreams:
Here are the results of `npm update` for the next release.


devinreams:
>Are there any bugfixes we need here? Given that we're pretty close to release (due in large part to 0.1.3 being delayed by Thanksgiving), I'd rather just defer this until we start work on 0.1.5

Agreed. I tried to get it in immediately after the 0.1.3 release but realize we have a short week. 

Removing from the 0.1.4 milestone...

devinreams:
>I'd rather just defer this until we start work on 0.1.5, unless this resolves some bustage we're having.

@jimporter How do you feel about re-reviewing our dependencies and updating them now?

devinreams:
I just force pushed this branch again with results of `update`.

devinreams:
>update -datastore to 0.1.0

@linuxwolf I may be seeing this wrong, but even though you set the URL for the datastore tag, I don't think npm knew what to do with it.

I got errors running `npm update` and also see this in the CI logs for `npm install`:

screen shot 2017-12-14 at 9 54 51 pm

So I tried just using the [GitHub URL](https://docs.npmjs.com/files/package.json#github-urls) syntax and it seems to work A-OK. How's it look to you?

linuxwolf:
@devinreams I was trying to avoid being too dependent on github URLs because it actually maps to a specific commit, and not the tag.

I suspect the issue with `npm update` is that it's very shallow on what it checks for updates, while `npm install` dives deeper -- if you ran `npm update` with the github URL from the start, you would have the same issue.

linuxwolf:
Also @devinreams, please attach text logs next time -- it's easier to do analysis over than screen shots.

devinreams:
>I was trying to avoid being too dependent on github URLs because it actually maps to a specific commit, and not the tag.

@linuxwolf Ah OK. 

Do you want me to revert the commit then? Or do we need to do something different?

linuxwolf:
@devinreams Personally I'm ambivalent.  The best solution is probably to publish to NPM; for now, whether it stays on the git branch or it goes through the tag archive is probably effectively the same difference.

What ought to happen is a note somewhere to do a deeper NPM update (e.g., `npm update --depth=9999` or event `rm -rf node_modules ; npm install`) if `package.json` and/or `package-lock.json` are updated.

pdehaan:
+1 to publishing **lockbox-datastore** to npm, and managing everything via semver.

devinreams:
Cool, now seems as good a time to drop it into npm. I’m all for it 👍

linuxwolf:
There's some questions in https://github.com/mozilla-lockbox/lockbox-datastore/issues/63 to address.  From this thread, it looks like the first two are basically answered (-:

linuxwolf:
Replaced with #472

PR #397, opened 2017-12-08
add external link widget

linuxwolf:
Created as a separate PR for now to (hopefully, but possibly vainly) help track any changes necessary here without conflicting with ongoing work in #351 .

### DIFF
--- a/src/webextension/widgets/link.js
+++ b/src/webextension/widgets/link.js
@@ -7,13 +7,19 @@ import React from "react";
 
 import styles from "./link.css";
 
-export function ExternalLink({onClick, children}) {
+export function ExternalLink({onClick, children, className, ...props}) {
+  const finalClassName = `${styles.external} ${className}`.trimRight();
   return (
-    <a className={styles.external} onClick={onClick}>{children}</a>
+    <a {...props} className={finalClassName} onClick={onClick}>{children}</a>
   );
 }
 
 ExternalLink.propTypes = {
-  onClick: PropTypes.func.isRequired,
   children: PropTypes.any,
+  className: PropTypes.string,
+  onClick: PropTypes.func.isRequired,
+};
+
+ExternalLink.defaultProps = {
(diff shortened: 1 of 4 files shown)


src/webextension/widgets/link.js
```diff
@@ -0,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import PropTypes from "prop-types";
+import React from "react";
+
+import styles from "./link.css";
+
+export function ExternalLink({onClick, children}) {
```
linuxwolf: defined it this way because I strongly suspect we'll have internal links in the near future, and thought this would help minimize the changes to consumers of this widget.

src/webextension/widgets/link.js
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <a className={styles.external} onClick={onClick}>{children}</a>
```
jimporter: Two small things. Let users of this component pass:

1. Extra `props` and apply them to the `<a>` tag: `<a {...props}>`
2. A `className` to add styles for a specific use of this component. This is the snippet I use elsewhere: `const finalClassName = `${styles.external} ${className}`.trimRight();`

src/webextension/widgets/link.js
```diff
@@ -0,0 +1,25 @@
...
+
+export function ExternalLink({onClick, children, className, ...props}) {
+  const finalClassName = `${styles.external} ${className}`.trimRight();
+  return (
+    <a {...props} className={finalClassName} onClick={onClick}>{children}</a>
+  );
+}
+
+ExternalLink.propTypes = {
+  children: PropTypes.any,
```
jimporter: This should be `PropTypes.node`.

PR #400, opened 2017-12-09
add integration test script runner

linuxwolf:
Adds "integration" to the package.json scripts, in order to run `tox`-based integration tests, with prerequisites.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,8 @@
     "run": "jpm run --addon-dir dist/ ${npm_config_jpm_runflags} --prefs jpm-prefs.json",
     "pretest": "npm run-script autolint",
     "test": "cross-env NODE_ENV=test karma start",
+    "preintegration": "cross-env ENABLE_DOORHANGER=1 npm run package",
+    "integration": "tox -r",
     "nsp": "nsp check",
     "codecov": "codecov",
     "clean": "git clean -fX ./dist ./*.xpi ./addon.env && npm run :clean:doc ",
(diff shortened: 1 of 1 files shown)


linuxwolf:
There is no hurry on getting this landed.  Merely a convenience for those of us working with integration tests a lot.

jrbenny35:
We should have a ```flake8``` as well as an ```integration test``` step. It would help with CI, to not have such a huge log if flake8 fails.

jimporter:
@jrbenny35 Would it make sense to merge the `flake8` and `eslint` checks into a single omni-lint step? (Plus `stylelint` one day.) I don't know if we need to break those out into separate things.

jrbenny35:
> Would it make sense to merge the flake8 and eslint checks into a single omni-lint step? (Plus stylelint one day.) I don't know if we need to break those out into separate things.

@jimporter I like it, if that would help. Just lint ALL THE THINGS.

devinreams:
>Would it make sense to merge the flake8 and eslint checks into a single omni-lint step?

@jimporter And if we do that, we take it out as a stage in the travis.yml right? Because it would run as part of test, if I understand correctly?

package.json
```diff
@@ -36,6 +36,8 @@
     "run": "jpm run --addon-dir dist/ ${npm_config_jpm_runflags} --prefs jpm-prefs.json",
     "pretest": "npm run-script autolint",
     "test": "cross-env NODE_ENV=test karma start",
+    "preintegration": "cross-env ENABLE_DOORHANGER=1 npm run package",
```
pdehaan: Just as a sanity check, do we still need this ENV var magic?
I thought we yanked all the `ENABLE_DOORHANGER` in PR #403; "Remove ENABLE_DOORHANGER code".

linuxwolf:
@jimporter I thought about changing around `test`, but figured that might be too much for this first go around.

package.json
linuxwolf: will remove; when first submitted this was still necessary.

linuxwolf:
rebased onto `master` and made the requested changes.

PR #403, opened 2017-12-12
Remove ENABLE_DOORHANGER code

### LINKED ISSUE #377: Enable doorhanger for production builds
After quick discussion with @jimporter and after running it by @sandysage, the product team does want the doorhanger enabled for the next release (0.1.4) as the benefits outweigh the cost. For example, it enables easier copy/paste to happen faster, even if it doesn't allow editing/adding.

This relies on Telemetry https://github.com/mozilla-lockbox/lockbox-extension/issues/358 being in place first.

Anything to add or discuss @sandysage @jimporter?

devinreams:
we won't ever want to have it not enabled from here forward, so removing the code entirely

closed #377 

@jimporter is this all there is to it? or more to do? feel free to jump on the branch directly if I missed anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ jobs:
         - pip install tox
       before_script:
         - npm install
-        - ENABLE_DOORHANGER=1 npm run package
+        - npm run package
       script: tox
       after_success: skip
       before_deploy: skip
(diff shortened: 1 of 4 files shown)


devinreams:
@jimporter looked at this over my shoulder yesterday and confirmed related code exists in just the 3 places (and the Travis config)

@linuxwolf would you mind being another set of eyes and make sure this builds/packages as expected for you, too?

PR #409, opened 2017-12-13
Fixes for QA findings before release

devinreams:
Addressing QA findings from @m8ttyB


devinreams:
Thanks @jimporter. Here's visual confirmation of the results after moving the overflow styles from item-fields to input:

screen shot 2017-12-13 at 11 29 42 am

PR #411, opened 2017-12-13
Fix feedback link on docs/index

pdehaan:
This was causing a 404 on the https://mozilla-lockbox.github.io/lockbox-extension/ site which was causing a bunch of other odd results from my link checker.


devinreams:
Embarrassed I missed this one. Thanks for looking out @pdehaan.

Issue #414, opened 2017-12-13
Odd blue background when selecting items from left side list


pdehaan:
### Steps to reproduce:
1. Click the "New Entry" button in the heaeder.
1. Click the "New Entry (no username)" list item from the left hand side.

### Actual results:

The list item's background is blue. The contrast makes it difficult to read the text.
Also, since it's the first/only item [and not saved yet], I see the placeholder onboarding text in the main content area instead of the add/edit form.

<img width="1280" alt="lockbox_entries_-_bleu" src="https://user-images.githubusercontent.com/557895/33963331-747ef8fa-e009-11e7-995a-6e203a711eaf.png">

Issue #297, opened 2017-11-07
Improve modal triggering for the management page


jimporter:
I'm not happy with how we decide to show modals for the management UI. The biggest issue is that our Redux action creators depend on our state, which means that state depends on actions and actions depend on state. It also relies on creating actions that essentially describe the state transition rather than the user action that occurred (Redux style prefers the latter).

This is something we really ought to fix before we start relying on it too heavily, but it's going to take a bit of research to figure out what the best way to do it is. Maybe `redux-saga` would be better than `redux-thunk`...?

jimporter:
We should also subdivide our action creators/reducers like https://github.com/erikras/ducks-modular-redux

PR #416, opened 2017-12-13
Don't fire an onChange event from <ScrollingList/> if the new item is already selected

jimporter:
This resolves - in a roundabout way - an issue with clicking the "new item" placeholder entry in the item list causing the editor widget to go away. It resolves a comment in #414.

Also note: this fixes the bug, but this was caused by a few bugs working together to screw everything up. The code for our Redux reducers is badong and I want to fix it (see #297).

Killing is Badong


jimporter:
Merged after a rebase.

Issue #402, opened 2017-12-11
Use :-moz-focusring instead of :focus


jimporter:
[It's better.](https://developer.mozilla.org/en-US/docs/Web/CSS/:-moz-focusring)

PR #418, opened 2017-12-14
Improve selected/focus state for the item list; resolves #414

### LINKED ISSUE #414: Odd blue background when selecting items from left side list

jimporter:
This improves the styling for the item list, especially where the styles are defined. Notable bits:

* `border-right` (now `border-inline-end` for RTL support) is only defined for the management UI, so it won't hurt the doorhanger
* Right border is now always shown, since the "no right border" thing didn't work great when we had a scrollbar. (Todo: we should probably talk to the Photon design team about whether overlay scrollbars should be a Photon thing.)
* Borders are 1px now (Zeplin shows them as 0.5px, but that's because it's a border on all sides, and we just do bottom/right.
* Focus state is improved, and should follow @changecourse's visual suggestions. We can (possibly should) continue to look at this to decide what we can do to improve it further.

### DIFF
--- a/src/webextension/list/components/item-list.css
+++ b/src/webextension/list/components/item-list.css
@@ -16,8 +16,8 @@
 }
 
 .item {
-  border-bottom: solid 0.5px #d7d7db;
-  border-right: solid 0.5px #d7d7db;
+  color: #0c0c0d;
+  border-bottom: 1px solid #d7d7db;
 }
 
 .item:nth-child(2n+1) {
--- a/src/webextension/list/components/item-list.js
+++ b/src/webextension/list/components/item-list.js
@@ -10,9 +10,10 @@ import ScrollingList from "../../widgets/scrolling-list";
 
 import styles from "./item-list.css";
 
-export default function ItemList({items, ...props}) {
+export default function ItemList({items, itemClassName, ...props}) {
+  const finalItemClassName = `${styles.item} ${itemClassName}`.trimRight();
   return (
-    <ScrollingList itemClassName={styles.item} data={items} {...props}>
+    <ScrollingList itemClassName={finalItemClassName} data={items} {...props}>
       {({id, title, username}) => {
         return (
           <ItemSummary id={id} title={title} username={username}/>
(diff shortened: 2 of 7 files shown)


devinreams:
The improved "click a new entry and it doesn't go back to the home screen" behavior is a big win here. I doubt people would do it but now that I know its there I certainly would like it to be fixed soon.

changecourse:
I'd like to request the following changes... screenshots also provided below.

- [ ] Change bg color of active entry to grey-50 (#737373) 
- [ ] Change text-color to white for selected entry
- [ ] Is it possible to have the active border only shown on click or keyboard toggle, as opposed to at all times? 

1 1 1 lockbox home - view entry

devinreams:
>Change bg color of active entry to grey-50 (#737373)
>Change text-color to white for selected entry

@changecourse That makes sense and looks good here after a quick test locally. No problem. 👍 

>Is it possible to have the active border only shown on click or keyboard toggle, as opposed to at all times?

If I understand correctly, you mean the "focus" state which, similar to our buttons and inputs, puts a blue border around the item in the scrolling list as seen on "Google" here:

screen shot 2017-12-14 at 9 15 10 pm

Assuming so, the border does show on click because it brings the item into focus. It also comes into focus if the user tabs through or presses the up/down arrows to navigate between items.

Focus remains in one place at a time (i.e.: the last thing that was focused) because nothing _removes_ focus other than.. something else _having_ the focus. So if you click "Google" then press <kbd>Tab</kbd> twice you're then "focused" on the "Copy" button next to the "Password" field. Or if you click on the search box, the focus is removed from the item and moved to that input.

So... that's a long way of saying: I don't think what you described is possible. As I understand it, "at all times" is how focus is intended to work. Does that answer your question?

@jimporter @linuxwolf did I mis-represent anything or any thoughts to add?

jimporter:
@devinreams @changecourse There's one small thing I'd like to add to that. With #402, there may be some cases where users won't see focus states if they're navigating via mouse (this is OS-dependent, so your mileage may vary).

changecourse:
Looks good, thanks @devinreams

devinreams:
@jimporter Thanks again for this. Do you mind resolving the test conflicts and merging?

I would go ahead and merge this myself for ya but I don't want to break anything. 🙃

jimporter:
Merged after a rebase

PR #423, opened 2017-12-14
update contributing URLs to point to right spots

### LINKED ISSUE #412: A few 404s on docs site
With #411 fixed/merged, the list of 404s shrinks considerably. Not sure what of these are worth fixing. My gut says "all", but I don't feel too passionate about /feed.xml and I don't know how much we care about stuff like SEO/RSS right meow.

lockox-404s

devinreams:
fixes #412 (partially)


devinreams:
Merging as-is since these are just minor doc URL changes.

PR #424, opened 2017-12-14
Allow running integration tests on Python 3.* instead of just 3.6

### LINKED ISSUE #420: Integration tests shouldn't require Python 3.6
I can't think of anything that would require Python 3.6 (as opposed to Python 3.5 or what-have-you) with our integration tests, so it'd be nice if we were more lenient about what versions of Python we allow. @jrbenny35 Any thoughts on this?

jimporter:
Resolves #420


jrbenny35:
LGTM!

PR #473, opened 2018-01-18
Fix new button; resolves #438

### LINKED ISSUE #438: Clicking "New entry" button while editing an entry will duplicate the entry
Steps to reproduce:
1. Edit an existing entry.
2. While in the "Edit Entry Details" mode, click the "New entry" button.

Result:
A new entry will be created in "Create New Entry" mode, but the form will be populated with the previous entry's values from "Edit Entry Details" mode.

Is this a feature or a bug? This behavior was surprising because the shift from "Edit Entry Details" mode to "Create New Entry" mode was subtle and it wasn't clear at first that I was now editing a different entry with the same values.

jimporter:


src/webextension/list/manage/components/edit-item-details.js
```diff
@@ -43,6 +43,14 @@ export default class EditItemDetails extends React.Component {
     this.state = {...props.fields};
   }
 
+  componentWillReceiveProps(nextProps) {
+    for (let i in this.props.fields) {
+      if (this.props.fields[i] !== nextProps.fields[i]) {
+        this.setState({[i]: nextProps.fields[i]});
+      }
+    }
```
jimporter: There's a problem with this that I'm not sure how to fix. The idea here is that when the component gets new props, it'll update any field whose *initial* value has been change (this is to prevent bogus prop updates from resetting form state). However, if you really do want to reset a field even when it's the same thing (e.g. two entries with the same email address), it won't reset.

I'll try to see if there's a cleaner way to reset the form, but if that doesn't work out, maybe I'll just drop the first commit here.

linuxwolf: I was going to ask about that in my review.

linuxwolf: There should be a comment here explaining why this is being done, if this code is retained.

jimporter:
@linuxwolf Ok, I rewrote the first commit to be a little more sensible, as well as adding a comment.

jimporter:
Merged after rebasing.

m8ttyB:
QA verified, the `New entry` button now disables while a user is editing an existing item or in the process of creating a new one.

PR #491, opened 2018-01-24
Implement a better copy-to-clipboard button; resolves #318

### LINKED ISSUE #318: Copy action verification to inform a user when a username/password is copied to the clipboard
When I copy my username or password, to populate a web form, I want verification that the item is copied to my clipboard, so that I can confirm that I'm pasting the right information into the field.

screen shot 2017-11-08 at 3 38 12 pm

## Acceptance Criteria
- Should have indication when either a username or a password is copied

## Expected behavior
- When "copy" action is clicked, next to either the username or the password, the "copied" feedback text appears
- "Copied" only shows next to either username or password, not both at once, to indicate which of the fields is copied to the clipboard
- The "copied" text is static and replaces the button when the item is copied to the clipboard
- Copied feedback shows for 3 seconds on a timer, then reverts back to the copy link
- Copied feedback is provided in both the mini-editor and the full-screen editor

jimporter:
@linuxwolf This ended up a bit larger than I'd anticipated, since there were a couple of things in `<ButtonStack/>` and the `react-copy-to-clipboard` package that were suboptimal and needed fixed now. Also, I tried to test the timeout for switching back to the "copy" button, but using sinon's fake timers broke a ton of our tests in mysterious ways... :(

@changecourse I didn't see a full spec for the interaction of the button in #318, so I just did my best to get something approximately correct. Let me know if you'd like anything to change. I also went with 2 seconds for the duration of the "Copied" label, since 3 seconds seemed really long to me. I can change it to 3 if you think the current patch is wrong, though.


sandysage:
@jimporter  what are some things that are missing from the issue to provide the information needed?

jimporter:
@sandysage I don't see a spec for what the focus state, hover state, or mousedown state should look like. I'm also not sure if the check mark should be an SVG or if Unicode is sufficient (I went with Unicode).

src/webextension/locales/en-US/widgets.ftl
```diff
@@ -1,10 +1,13 @@
+copy-to-clipboard-button = Copy
+copy-to-clipboard-copied = ✔ Copied
```
linuxwolf: Just explicitly noting what I hope is obvious: how this unicode renders depends on the font in use.  If we change the font, it can (read: likely will) change how this renders.

It will still be a checkmark (I hope), but the shape can be different in noticeable ways.

I am perfectly happy to use unicode instead of images wherever possible.

linuxwolf:
@jimporter could you elaborate how and where things broke with fake timers?  Not saying we try to fix all those things now, but it would help figure it out.

jimporter:
@linuxwolf Every test run after I injected the fake timer (even though I cleaned it up in `afterEach`) failed. The first one failed with `performance.mark is not a function`, which is weird, since it should exist in a browser context...

Now that I think about it, it might be a jsdom issue and I could work around it with the `mountIntoDOM` function that makes a *real* element.

jimporter:
@linuxwolf Every test run after I injected the fake timer (even though I cleaned it up in `afterEach`) failed. The first one failed with `performance.mark is not a function`, which is weird, since it should exist in a browser context...

Now that I think about it, it might be a jsdom issue and I could work around it with the `mountIntoDOM` function that makes a *real* element.

src/webextension/locales/en-US/widgets.ftl
jimporter: Yeah, we *should* know what all the fonts are (and Firefox is pretty good about falling back to another font if the current one doesn't have that glyph), so it's not a complete unknown. Still, I'll leave it up to @changecourse to decide what we should do here. Using Unicode does have the benefit of being a fair bit simpler though...

jimporter:
@linuxwolf Ok, well I just looked into how sinon does this and wrote my own extremely-simple mock for `setTimeout`.

linuxwolf:
@sandysage in action!
copy-confirmed

m8ttyB:
@changecourse Does the UX and UI match your needs?

linuxwolf:
**NOTE** per discussed release process, awaiting `pi-verified` before merging to master.

jimporter:
@linuxwolf Oh, that's ok. I was just rebasing and forgot to click merge.

PR #500, opened 2018-01-26
Remove unused <ErrorMessage/>; resolves #240

### LINKED ISSUE #240: Be careful with withLocalization()/getString()

jimporter:
We're no longer using `<ErrorMessage/>`, and it uses `getString()` (which we should do our best to avoid). It also doesn't have any tests, so let's remove it!


jimporter:
No PI needed here, since it's unused code and won't affect the build.

PR #506, opened 2018-01-29
Implement light vs. full signout

### LINKED ISSUE #149: Allow user to "Lock" editor via "Sign Out" to prevent unwanted access to Lockbox data

linuxwolf:
Resolves #149


m8ttyB:
@linuxwolf Is there a way to hit this from the browser? I don't quite understand how to verify it's behavior.

m8ttyB:
Capturing some testing notes:
* Thank you for pairing on testing this, I've captured notes from our session and translated them into test cases that will live in TestRail
* We identified a broadcast event that needs to be added as well as well as a telemetry event.

PR #507, opened 2018-01-30
Provide documentation about how to post PRs

jimporter:


docs/contributing.md
```diff
@@ -6,24 +6,56 @@ The following are guidelines for contributing to this project.
...
+Before you get started writing code, be sure what you plan to work on is
+something we'll be able to accept. The easiest way is to look through out list
+of [good first issues][good-first-issues] and find something that sounds
+interesting. If there's something else you'd like to work on, just add a comment
+in the relevant issue and we'll be happy to discuss your plans.
+
+Once you have a patch ready, submit a pull request and request a review from
+**@mozilla-lockbox/engineering**. From there, we'll guide you through the review
+process to the eventual landing of your code!
+
```
linuxwolf: As agreed in a team discussion on 2018-01-26, there should be guidance on when an external contributor can expect to hear something from this team.  Something along the lines of the following (@sandysage @kfontecilla can correct me if mistaken):

```
Someone from this team will make the best effort to review or comment on your changes within two working days.
```

docs/contributing.md
```diff
@@ -6,24 +6,56 @@ The following are guidelines for contributing to this project.
...
 
-Please keep in mind, by opening an issue we provide no guarantee the enhancement will be implemented.
+Please keep in mind, by opening an issue we provide no guarantee the enhancement
+will be implemented.
+
+## How to Contribute Code
+
+Before you get started writing code, be sure what you plan to work on is
+something we'll be able to accept. The easiest way is to look through out list
+of [good first issues][good-first-issues] and find something that sounds
```
devinreams: Is this “first issues” link defined? I don’t see it anywhere in this doc...

jimporter:
> Why are line wrappings hard-coded into the document?

I'd say hard-wrapping is the most tool-agnostic way to do things, since it should be readable with pretty much anything, including something as simple as `cat`. Since one of the main pluses of Markdown is that it's human-readable, I think it's best to make it as easy as possible for people to read it easily with whatever tool they have on-hand.

docs/contributing.md
jimporter: I think I forgot that. I'll add it in.

docs/contributing.md
jimporter: Do we want to make a public commitment to that? I thought the plan was that the 2-day deadline was just for us to know about. I'm just hesitant to say we'll do this when it might be hard to guarantee for outside contributions (e.g. if you and I are both out for a couple days); for internal stuff, we'd know who's out, but for external patches, the contributor probably wouldn't.

That said, I don't have a strong feeling on the matter, so if everyone else thinks we should say this in the docs, I'm ok with it.

linuxwolf: The plan we discussed was to add it to public documentation, and my recollection is we had consens on "best effort" (see my sample above).  Best effort is not a guarantee.

linuxwolf:
To use and expand on your terms: this is 80-column no-wrap "zealotry".

The number of tools that _don't_ soft wrap by default are code writing tools, and even there the (anecdotal) majority treat documentation formats (e.g., HTML, markdown, RST) differently.  Even `cat` (really, the terminal/console/tty onto which it's IO-bound) and `less` (or `more`) soft-wrap appropriately for most people.  But when you're editing such documents, tools that keep hard-wraps properly in place are far rarer.  We hope this particular document doesn't go through this, but the precedent for editing documentation content ought to favor the editor when so many rendering environments render how a person expects.

jimporter:
Ok, well at the very least, I'll add a `.dir-locals.el` to our docs dir so I can have visual line wrapping in emacs.

docs/contributing.md
pdehaan: Ref: https://github.com/mozilla-lockbox/lockbox-extension/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

m8ttyB:
Doing a quick review of content I'm okay with the document. Giving my 👍 🌮

jimporter:
This should be ready for another review...

linuxwolf:
@devinreams this looks good from my perspective.  Are you happy with it?

devinreams:
Thanks @jimporter and @linuxwolf for thinking about and writing this out for others 🌊🚢

PR #546, opened 2018-02-14
pin and ignore to eslint-plugin-mozilla@0.6.0

linuxwolf:
This both pins us to `eslint-plugin-mozilla` version 0.6.0 and ignores future updates (for a bit).  It appears this plugin is broken (and some out-of-band follow up on that is needed), but this prevents some extraneous noise.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1038,6 +1038,12 @@
         }
       }
     },
+    "babel-helper-evaluate-path": {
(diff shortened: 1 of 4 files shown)


linuxwolf:
@mozilla-lockbox/product-integrity for verification, if you can run an older version of FF (between 57 and 58), that would help verify this is working as intended.

PR #547, opened 2018-02-15
(WIP) Refresh OAuth Access Tokens

### LINKED ISSUE #487: Implement "light" vs. "full" account signout
Signing out of an account usually should change the "mode" from "authenticated" to "unauthenticated", but still allow for some authorized behaviors:

* retrieving remote entries
* noting the user's FxA `uid` for telemetry and other purposes
* remember (in memory) the user's display name and avatar image (if available).

Additionally, sign out can perform a "full" or complete (local) reset of account details, effectively resetting the account mode to "guest".

Related to:
* #209
* #384
* #149

linuxwolf:
Resolves #487 

Implements the token refresh (and, as a side effect, user profile info) method.

This is not wired up to anything as yet.

### DIFF
--- a/src/webextension/background/accounts/index.js
+++ b/src/webextension/background/accounts/index.js
@@ -190,29 +190,25 @@ export class Account {
 
   async token() {
     // always return null if user is GUEST
-    if (this.mode() === GUEST) {
-      return null;
+    if (this.mode === GUEST) {
+      // XXXX: use DataStoreError
+      throw new Error("AUTH: requires FxA");
     }
 
     // check if token present / unexpired / valid
-    await this.updateUserInfo();
-    if (!this.info || !this.info.access_token) {
+    let info = await this.updateUserInfo();
+    if (!info || !info.access_token) {
       // refresh
-      const { refresh_token } = this.info || {};
-      if (!refresh_token) {
-        throw new Error("AUTH: cannot refresh token");
-      }
-
       await this.updateAccessToken();
-      await this.updateUserInfo();
+      info = await this.updateUserInfo();
     }
 
-    if (!this.info || !this.info.access_token) {
+    if (!info || !info.access_token) {
       // XXXX: use DataStoreError
       throw new Error("AUTH: no access token");
     }
 
-    return this.info.access_token;
+    return info.access_token;
   }
 
 
@@ -223,8 +219,8 @@ export class Account {
     if (!params) {
       // assume "refresh_token" exchange
       if (!info.refresh_token) {
-        // TODO: use categorized DataStoreErrors
-        throw new Error("AUTH: refresh token required");
+        // XXXX: use DataStoreError
+        throw new Error("AUTH: no refresh token");
       }
 
       params = {
@@ -240,7 +236,7 @@ export class Account {
         "content-type": "application/json",
       },
       cache: "no-cache",
-      body: JSON.stringify(params),
+      body: params,
     };
     const oauthInfo = await fetchFromEndPoint("token", cfg.token_endpoint, request);
     let keys = info.keys || new Map();
--- a/src/webextension/background/accounts/index.js
+++ b/src/webextension/background/accounts/index.js
@@ -66,6 +66,52 @@ async function fetchFromEndPoint(name, url, request) {
   return body;
 }
 
+async function fetchAccessToken(url, params) {
+  const request = {
+    method: "post",
+    headers: {
+      "content-type": "application/json",
+    },
+    cache: "no-cache",
+    body: JSON.stringify(params),
+  };
+  return fetchFromEndPoint("token", url, request);
+}
+
+async function validateAccessToken(acct) {
+  const cfg = configs[acct.config];
+
+  const { access_token, expires_at } = this.info || {};
+  if (!access_token || !expires_at || Date.now() > (expires_at * 1000)) {
+    // needs a new access token
+    return null;
+  }
+
+  const url = cfg.token_endpoint;
+  const request = {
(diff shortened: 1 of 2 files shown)


linuxwolf:
@mozilla-lockbox/product-integrity 

As written, the `chore` does not require any user interaction today.  I'm not sure how much verification is possible outside tests, until more sync logic is implemented.

m8ttyB:
@linuxwolf agreed, I'm not terribly concerned about testing this on the PI side until this gets wired into a flow that a user can interact with.

Peer review as well as the accompanying tests in the PR should be plenty.

src/webextension/background/accounts/index.js
```diff
@@ -228,6 +187,118 @@ export class Account {
...
+        key = await jose.JWK.asKey(key);
+        keys.set(name, key);
+      });
+      await Promise.all(pending);
+    }
+
+    this.info = info = {
+      ...info,
+      access_token: oauthInfo.access_token,
+      expires_at: 0 | (Date.now() / 1000) + oauthInfo.expires_in,
```
jimporter: This `0 | ` looks new; you're truncating to an int right? Could we use `Math.trunc` or `Math.floor` instead? Those would be clearer (`Math.trunc` should have identical behavior, but `Math.floor` is probably clearer; it only matters if the arg is negative, which it shouldn't be). If there's a particular reason those wouldn't work, then feel free to leave it the way it is...

test/unit/background/accounts-test.js
```diff
@@ -108,14 +108,14 @@ describe("background > accounts", () => {
       uid: "1234",
       access_token: "KhDtmS0a98vx6fe0HB0XhrtXEuYtB6nDF6aC-rwbufnYvQDgTnvxzZlFyHjB5fcF95AGi2TysUUyXBbprHIQ9g",
       id_token: "eyJhbGciOiJSUzI1NiIsImtpZCI6IjBDWTBJb3RVVHBtWDFDbXJqMWNwcTB6aFBkd1NtalJSaWlJNzduMmMtMFkifQ.eyJ1aWQiOiIxMjM0IiwiaXNzIjoic2NvcGVkLWtleXMifQ.pX8I1LqCD849Gp0TzKcS6LM_fko0gc7wkSzBgPaxFDyF8AZrWn9-HTRoW-9YIuHujLzldbI1k34VeSHNM85vkPjm_AxbBKuXEiVJQdcCAxNjbSQmM1dOX6kZKwN4oDu8X4BB3CwQq5eXioYYiPur149O_I2bhFDuMBtQBoQosZtOScuKliXcURuWEwhYcnHe8axit0fQ0vd1FOJK3300hccqcZNoHGXrSVj42mdo_aSREOcwSUP4i0r0aCfJqnxai43uy1C5l54mSN1KzqGeasx60lWPU-Jm3gPm_2CXRWbfWxF3-OnxhMhSQiS90kefX81H03ZYVShDutsx55d0tQ",
+      expires_at: 0 | (Date.now() / 1000) + 1209600,
```
jimporter: As above, `Math.trunc` or `Math.floor` would be clearer.

src/webextension/background/accounts/index.js
linuxwolf: it is for int truncation, and the habit I got into was using this over `Math.floor` because of a very significant (order of magnitude) performance difference in the past.  I can switch them to floor, though.

linuxwolf:
@mozilla-lockbox/product-integrity @jimporter all comments addressed, plus taking better steps to actually _remember_ the refreshed info.

test/unit/background/accounts-test.js
```diff
@@ -103,24 +94,27 @@ describe("background > accounts", () => {
   });
 
   describe("Account", () => {
-    let acct;
+    let acct, storage;
     const unauthedInfo = {
       uid: "1234",
       access_token: "KhDtmS0a98vx6fe0HB0XhrtXEuYtB6nDF6aC-rwbufnYvQDgTnvxzZlFyHjB5fcF95AGi2TysUUyXBbprHIQ9g",
       id_token: "eyJhbGciOiJSUzI1NiIsImtpZCI6IjBDWTBJb3RVVHBtWDFDbXJqMWNwcTB6aFBkd1NtalJSaWlJNzduMmMtMFkifQ.eyJ1aWQiOiIxMjM0IiwiaXNzIjoic2NvcGVkLWtleXMifQ.pX8I1LqCD849Gp0TzKcS6LM_fko0gc7wkSzBgPaxFDyF8AZrWn9-HTRoW-9YIuHujLzldbI1k34VeSHNM85vkPjm_AxbBKuXEiVJQdcCAxNjbSQmM1dOX6kZKwN4oDu8X4BB3CwQq5eXioYYiPur149O_I2bhFDuMBtQBoQosZtOScuKliXcURuWEwhYcnHe8axit0fQ0vd1FOJK3300hccqcZNoHGXrSVj42mdo_aSREOcwSUP4i0r0aCfJqnxai43uy1C5l54mSN1KzqGeasx60lWPU-Jm3gPm_2CXRWbfWxF3-OnxhMhSQiS90kefX81H03ZYVShDutsx55d0tQ",
+      expires_at: Math.floor(Date.now() / 1000) + 1209600,
```
pdehaan: Not sure if we want to add a comment (or `const`) about the seemingly magic "1,209,600" number represents (my best guess was 2 weeks in seconds).

linuxwolf: @pdehaan for testing, the value is really arbitrary.  Personally, I'm inclined to let it be arbitrary for now.

Issue #442, opened 2018-01-04
Display dropdown menu and links to give users access to account management features


sandysage:
Implement a dropdown menu giving the user the ability to:

* review and update their FxA account
* sign out of Lockbox
* access options/preferences for Lockbox

<img width="249" alt="image" src="https://user-images.githubusercontent.com/757401/34845624-b98a3bbe-f6d2-11e7-9bd6-470e7afa2e85.png">

## Acceptance criteria
- [x] Should have a way to navigate it on a keyboard

## To do:
- [ ] needs discussion for keyboard accessibility to manage any changes to be made in its own story
- [ ] display dropdown menu when clicking on "user info" on full-screen
- [ ] open https://accounts.firefox.com/settings when clicking "Account"
- [ ] open Options when clicking "Preferences" menu item
- [ ] sign out when clicking "Sign out" menu item (blocked by #149)

sandysage:
- [ ] needs discussion for keyboard accessibility to manage any changes to be made in its own story

PR #549, opened 2018-02-15
Styling adjustments for the account dropdown.

### LINKED ISSUE #524: Add an account dropdown (additional styling considerations)
Placeholder story for additional work on the account dropdown, including...

- [x] Sizing of list items
- [x] Alignment of iconography to list items
- [x] Default width of dropdown relative to header
- [x] hover and click states for list items
- [ ] animation of dropdown (in and out)
- [x] cursor style (default vs pointer) for links

A potential design consideration for future DR work, not a priority for now...

nickbrandt:
Resolves #524 

- Sizing of list items
- Alignment of iconography
- Min-width of dropdown
- Added hover and click states
- Cursor to pointer
- Adjusted svg icon files for proper sizing


jrbenny35:
Could we get maybe a screen recording of how it looks?

linuxwolf:
here is what it looks like for me:

524-acct-dropdown

src/webextension/list/manage/components/account-summary.css
```diff
@@ -65,13 +65,13 @@
   /* XXX: This positioning is a hack and assumes that the <AccountSummary/> ends
      up in a particular spot in the UI. It would be nice if we could make this
      more flexible in the future. */
-  top: 50px;
-  right: 2px;
-  min-width: 130px;
-  padding: 8px 12px;
-  border: 1px solid #b1b1b3;
+  top: 60px;
+  right: 16px;
```
jimporter: Why these changes? I intentionally set the top/right position to match the visuals in #442, so that the caret beside the username lines up with the arrow in the "speech bubble" of the dropdown.

src/webextension/list/manage/components/account-summary.css
```diff
@@ -124,18 +145,20 @@
   height: 16px;
   background-repeat: no-repeat;
   background-size: 16px 16px;
-  margin-inline-end: 1ch;
+  margin-right: 8px;
```
jimporter: Please continue to use `margin-inline-end` here; it'll look weird otherwise in RTL (granted, the popup will probably end up in the wrong place entirely in RTL, but that's a separate issue).

src/webextension/list/manage/components/account-summary.css
```diff
@@ -109,12 +109,33 @@
...
+  text-decoration: none;
+}
+
+.dropdown-link:hover {
+  background-color: #E5EFFA;
+}
+
+.dropdown-link:active {
+  background-color: #0060DF;
+  color: #fff;
```
jimporter: Nit: as a general rule, we use the full 6-digit hex value (`#ffffff`).

src/webextension/list/manage/components/account-summary.css
```diff
@@ -65,13 +65,13 @@
...
-  top: 50px;
-  right: 2px;
-  min-width: 130px;
-  padding: 8px 12px;
-  border: 1px solid #b1b1b3;
+  top: 60px;
+  right: 16px;
+  min-width: 150px;
+  padding: 8px 0;
+  border: 1px solid #D4D4D5;
```
jimporter: Nit: as @linuxwolf said, lowercase here.

src/webextension/list/manage/components/account-summary.css
```diff
@@ -65,13 +65,13 @@
...
   border-radius: 6px;
-  box-shadow: 0 2px 8px rgba(12, 12, 13, 0.1);
+  box-shadow: 0 4px 10px rgba(0, 0, 0, 0.2);
```
jimporter: Where's this color coming from? The Photon spec uses the previous value: https://design.firefox.com/photon/patterns/shadows.html

src/webextension/list/manage/components/account-summary.css
```diff
@@ -109,12 +109,33 @@
...
-}
-
 .dropdown-link {
   color: #0c0c0d;
+  display: block;
+  width: 100%;
+  text-align: left;
+  padding: 8px 16px;
+  border-radius: 0;
+  transition: background .15s ease;
```
jimporter: Where's this coming from? From what I can see, the Zeplin mockup doesn't have hover states*, and moreover, the Photon spec doesn't appear to animate hover states for any of the widgets in the spec. If we *did* want to use an animation here (and I'm not sure we do), I think we'd also want to use the Photon animation curve: https://design.firefox.com/photon/motion/duration-and-easing.html

\* One of the many issues I've had with Zeplin...

jimporter: Also, a minor point that you don't have to fix, but we're making these widgets look a lot like buttons. I wonder if we should switch to a `<Button/>` element in the JS (which would give us the "button" ARIA role for free). @linuxwolf Thoughts?

src/webextension/list/manage/components/account-summary.css
```diff
@@ -124,18 +145,20 @@
   height: 16px;
   background-repeat: no-repeat;
   background-size: 16px 16px;
-  margin-inline-end: 1ch;
+  margin-right: 8px;
   vertical-align: middle;
+  margin-top: -2px;
```
jimporter: This seems a bit brittle to me, but I'm not sure if there's a better way to make it look "right"...

src/webextension/list/manage/components/account-summary.css
```diff
@@ -124,18 +145,20 @@
   height: 16px;
   background-repeat: no-repeat;
   background-size: 16px 16px;
```
jimporter: If we're switching from `background-image` to `mask-image` below, we don't need these two lines anymore, right?

src/webextension/list/manage/components/account-summary.css
```diff
@@ -109,12 +109,33 @@
...
 }
 
-.dropdown > ul > li {
-  padding: 8px;
-}
-
 .dropdown-link {
   color: #0c0c0d;
+  display: block;
+  width: 100%;
```
jimporter: With this and the other sizing changes, I think we'll want to revisit what the focus-ring should look like for these elements. It looks kind of broken right now... (to see them in action, just open the dropdown and hit tab to cycle the focus).

src/webextension/list/manage/components/account-summary.css
linuxwolf: the latest [mockups](https://zpl.io/25gqr00) remove the chevron and reposition the avatar, which then makes these "speech bubble" alignment changes more understandable.

It would be helpful if this PR included those, but they also go beyond simply CSS changes.  @nickbrandt would you like some help on that?

src/webextension/list/manage/components/account-summary.css
nickbrandt: @jimporter I agree, wasn't able to get them to line up perfectly unless the text was within a span. Or could use position relative to move this up but figured it wasn't much difference. Let me know if you want it updated to another way.

src/webextension/list/manage/components/account-summary.css
linuxwolf: @jimporter @nickbrandt  @kimberlythegeek 

It's not clear what the full benefits of `role="button"` over `role="link"` are in this case.  I'm inclined to leave the PR as is for now, and address in the future as needed.

nickbrandt:
Per discussion with @changecourse, the focus state for the dropdown items has been updated to reflect the same as the hover state (similar to how Firefox handles dropdown states for things like Library in the browser). Also reverted the icons to use background-image again as we ended up deciding not to invert the list items on press state.

src/webextension/list/manage/components/account-summary.js
```diff
@@ -18,8 +18,8 @@ export function AccountSummaryButton({displayName, avatar, onClick}) {
 
   return (
     <button className={styles.button} onClick={onClick}>
-      <span className={styles.avatar}><img src={avatar}/></span>
       <span className={styles.displayName}>{displayName}</span>
+      <span className={styles.avatar}><img src={avatar}/></span>
```
linuxwolf: Moving the avatar from first to last requires some changes to the unit tests to account for the change in position:

*  [account-summary-test.js line 26](https://github.com/mozilla-lockbox/lockbox-extension/blob/c87cb321abc383e295f491a1b261488555ad2423/test/unit/list/manage/components/account-summary-test.js#L26]) -- `expect(wrapper.find("span").at(1))` should be `expect(wrapper.find("span").at(0))`
* [account-summary-test.js line 115](https://github.com/mozilla-lockbox/lockbox-extension/blob/c87cb321abc383e295f491a1b261488555ad2423/test/unit/list/manage/components/account-summary-test.js#L115) -- `expect(wrapper.find("span").at(1))` should be `expect(wrapper.find("span").at(0))`
* [current-account-summary-test.js line 47](https://github.com/mozilla-lockbox/lockbox-extension/blob/c87cb321abc383e295f491a1b261488555ad2423/test/unit/list/manage/containers/current-account-summary-test.js#L47) -- `expect(wrapper.find("span").at(1))` should be `expect(wrapper.find("span").at(0))`

kimberlythegeek:
Did a code review, and also verified that keyboard interaction still works. 👍

linuxwolf:
Merging as all review comments have been addressed, automated tests pass, and PI has verified.

PR #552, opened 2018-02-19
Add Support For Telemetry Scalars; Update Metrics.md

irrationalagent:
Here's my attempt at adding scalar recording support for telemetry. Scalars allow us to record fixed quantities or string values, rather than events.

When a scalar is set, it gets submitted under `processes.dynamic.scalars` in the main ping. 

In this PR I added recording for `datastoreCount` which is the number of items in the datastore:
screen shot 2018-02-19 at 12 26 53 pm

I will probably need some feedback on where the best place to record this is. For now, I slid it into background/message-ports.js in the part where the list of datastore items are retrieved for display. 

This means that if the user never displays their item list, the scalar won't get updated. That isn't ideal, but I was unsure of the safety of calling `openDataStore()` outside of that context, just for setting the scalar. If reviewers think that's OK to do, then I can make a change. 

Also, things I'm not sure about:

1. I guess I don't fully understand the need for the`proxy_telemetry_event` message in src/webextension/telemetry.js ... but I added `proxy_telemetry_scalar` in any case for consistency.

2. I don't really understand the implementation of unit tests at all, so what I added there is simply designed to be parallel to the event recording case, when the event is already registered. I'll probably need some help here :)

Finally, I added updates to metrics.md to:
- Describe how our measures help us track our KPIs
- Briefly describe how and what scalars are recorded
- Add a brief list of possible iOS events (maybe doesn't need to be in extension repo?)
- Remove outdated things

@linuxwolf I added you as a reviewer in Jim's absence. Not the end of the world if we can't get this merged before the end of the sprint.


irrationalagent:
well the travis build failed predictably on my unit test fumbling.. will have to get some guidance there

src/bootstrap.js
```diff
@@ -162,9 +162,28 @@ function startup({webExtension}, reason) {
...
+      console.log("telemetry scalar already registered; skipping registration");
+    } else {
+      throw e;
+    }
+  }
+
   webExtension.startup().then(({browser}) => {
     Services.telemetry.recordEvent(TELEMETRY_CATEGORY, "startup",
                                    "webextension");
+    // Services.telemetry.scalarSet(TELEMETRY_CATEGORY + ".datastoreCount", 50);
```
linuxwolf: is there a need for keeping this comment?  If not, we should remove it so we don't have "dead code".

src/webextension/background/message-ports.js
```diff
@@ -186,6 +188,9 @@ export default function initializeMessagePorts() {
     case "proxy_telemetry_event":
       return telemetry.recordEvent(message.method, message.object,
                                    message.extra);
+    case "proxy_telemetry_scalar":
```
linuxwolf: I don't see where this is actually used.

I think you included this for the sake of completeness.  That's ok if  if we think it's likely other scalars will need recording from the front-end (e.g., full editor view, doorhanger, etc).

If we keep this, it will need a test in `test/unit/background/message-ports-test.js`; otherwise we can remove it for now.

src/webextension/telemetry.js
```diff
@@ -7,3 +7,9 @@ export async function recordEvent(method, object, extra) {
     type: "proxy_telemetry_event", method, object, extra,
   });
 }
+
+export async function setScalar(name, value) {
```
linuxwolf: As with the handler in `message-ports.js`, this method is not used anywhere.

I think you included this out of completeness, which is ok if we see other front-end driven scalars coming in; in which case there should be a test.  Otherwise, it's fine to just remove this (and the `"proxy_telemetry_scalar"` handler) for now.

src/bootstrap.js
```diff
@@ -173,6 +191,12 @@ function startup({webExtension}, reason) {
           message.extra || null
         );
         respond({});
+        break;
+      case "telemetry_scalar":
+        Services.telemetry.scalarSet(
+          TELEMETRY_CATEGORY + message.name, message.value
```
linuxwolf: I should have called this out in the first review (sorry!):

I think it would be better for the scalar names to be "plain" (e.g., no leading `.`).  To that end, this line should change to something like:

```javascript
`${TELEMETRY_CATEGORY}.${message.name}`
```

irrationalagent:
OK, I think I addressed the changes. tests seem to be passing now

linuxwolf:
@mozilla-lockbox/product-integrity do you think you need to look at this?  Most likely the answer is "yes" ...

irrationalagent:
@m8ttyB I added a description of the scalar in metrics.md. In order to check if the scalar is being recorded, go to about:telemetry, click scalars, then select "dynamic" from the dropdown (see screenshot in op). If you want to chat about this one on one, let me know

PR #558, opened 2018-02-21
Update Version and Release Notes to 0.1.7-alpha

linuxwolf:


linuxwolf:
@changecourse or @sandysage, are you happy with the release notes herein?

changecourse:
Fine by me.

sandysage:
I propose keeping with the same tone as we have with other releases where we focusing on the value the user gets from the change(s) -- "you get this great thing." So in this instance, I would rewrite the new update as:

`The visual design and interaction of account actions dropdown, in the full editor view, better aligns with Photon standards to provide you with a closer brand experience with Firefox`

Or something like that.

PR #560, opened 2018-02-21
Updated text color of entry list empty state for better contrast

### LINKED ISSUE #554: Insufficient color contrast on empty Lockbox Entries list
The text "When you add an Entry, it automatically shows up here." has insufficient color contrast against its background.

Background: `#e0e0e3`
Text: `#737373`
Expected Contrast Ratio: `4.5:1`
Contrast Ratio: `3.58`

Text should be larger, or contrast should be higher. 
Some recommended text colors: `#555555`, `#545454`, `#3e3e3e`

nickbrandt:
Resolved #554 
Changed color from #737373 to #4a4a4f


linuxwolf:
old: 
image

new: 
image

PR #567, opened 2018-02-23
Create/Edit Entry Details Styling Adjustments

### LINKED ISSUE #479: DR: Refine the ‘create’ and ‘edit’ entry forms to better align to photon
*Adjustments to the form fields of 'Create Entry' and 'Edit Entry'*

**Invision (Create): https://mozilla.invisionapp.com/share/W3EU997YN#/263643956_1-1-0_Lockbox_Home_-_New_Entry_-_Unfilled**

**Invision (Edit): https://mozilla.invisionapp.com/share/W3EU997YN#/262163359_1-1-2_Lockbox_Home_-_Edit_Entry**

## Tasks ##
- [x] Form field vertical padding between fields in both the read-only and edit mode (20px)
Button widths (132px for primary, 80px for secondary)
- [x] Change ‘site address’ to ‘website address’ throughout
- [x] Margins for fields within content window (40/40 from left and top)
- [x] Padding/placement of Discard Changes & Delete dialog to match comps
- [x] Red delete button

nickbrandt:
Resolves #479 

- Updated padding/sizing of fields and container.
- Fixed labels and placeholder text to match new designs
- Added show/hide icons to replace text links for password field
- Reordered fields to match latest designs
- Removed monospace class from username
- Created new button style for red/critical buttons (e.g. Delete)
- Added min-width to buttons
- Updated padding/sizing of dialog box and content within to match designs
- Made global change of **box-sizing: border-box** to make sizing elements easier, as well as updated areas that this effected


src/webextension/icons/show.svg
```diff
@@ -0,0 +1,18 @@
...
+            <g id="ic_remove_red_eye">
+                <rect id="bounding-box" x="0" y="0" width="16" height="16"></rect>
+                <g id="Icon-24px" transform="translate(0.000000, 2.000000)">
+                    <path d="M1,5.95041322 C1,8.20254695 4.73761678,10.9008264 8,10.9008264 C11.2623832,10.9008264 15,8.20254695 15,5.95041322 C15,3.6982795 11.2623832,1 8,1 C4.73761678,1 1,3.6982795 1,5.95041322 Z" id="Shape" stroke-opacity="0.6" stroke="#0C0C0D" stroke-width="2"></path>
+                    <circle id="Oval" fill-opacity="0.6" fill="#0C0C0D" cx="8" cy="6" r="3"></circle>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
jimporter: Let's run this through `svgo` to eliminate the extraneous bits.

src/webextension/list/manage/components/app.css
```diff
@@ -2,6 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+html {
+  box-sizing: border-box;
```
jimporter: I don't have a strong opinion either way on whether we should do this (though I lean softly towards not doing it as it makes our modular CSS more brittle), but I don't think now is the right time for it. The biggest issue is that this is only being applied in the management tab, so the `width`/`height` of things used in multiple views will be inconsistent (stuff in the `widgets/` dir is the most common example).

If we do want to change what box-sizing we use, it should be its own issue so that we cover everything and have an opportunity to compare things so that we can be sure there are no accidental visual changes.

src/webextension/list/manage/components/item-details.css
```diff
@@ -3,16 +3,47 @@
...
+
+input[name="password"] + div > span:first-child > button:before {
+	background-image: url(/icons/show.svg);
+}
+
+input[name="password"] + div > span:last-child > button:before {
+	background-image: url(/icons/hide.svg);
+}
+
+textarea {
```
jimporter: As mentioned above, we use CSS class names here instead of tag names. There isn't a class name set on the textarea element in `item-fields.js`, but you can just set className to `notes` and then update this.

src/webextension/widgets/button.js
```diff
@@ -11,6 +11,7 @@ const THEME_CLASS_NAME = {
   primary: `${styles.primaryTheme}`,
   normal: `${styles.normalTheme}`,
   ghost: `${styles.ghostTheme}`,
+  critical: `${styles.criticalTheme}`,
```
jimporter: A minor issue, but I generally see this called "danger" in Mozilla source, instead of "critical". Let's go with that so other Moz folks have an easier time understanding our conventions.

src/webextension/widgets/dialog-box.js
```diff
@@ -36,8 +36,8 @@ export default class DialogBox extends React.Component {
             let primaryProps = {};
             if (i === 0) {
               primaryProps = {
-                theme: "primary",
-                ref: (element) => this._primaryButton = element,
+                theme: "critical",
```
jimporter: We shouldn't change this to `critical` (or `danger`) in all cases; some modals will want less-scary UI (in particular, I think the "cancel editing" dialog would want a blue button instead of a red one). I think we'd want to pass in `danger` as a prop to the `<DialogBox/>` so people who want a dangerous dialog can just pass `danger={true}` to it or something.

src/webextension/widgets/toolbar.css
```diff
@@ -10,7 +10,7 @@
 }
 
 .toolbar > * + * {
-  margin-inline-start: 5ch;
+  margin-inline-start: 2.5ch;
```
jimporter: Wasn't there another PR I reviewed where we ended up changing this to px? I think we can drop this change.

src/webextension/list/components/item-fields.js
```diff
@@ -28,26 +28,26 @@ const fieldsPropTypes = PropTypes.shape({
...
-          <LabelText>tITLe</LabelText>
-        </Localized>
-        <FieldText data-name="title">{fields.title}</FieldText>
-      </div>
       <div className={styles.field}>
         <Localized id="item-fields-username">
           <LabelText>uSERNAMe</LabelText>
         </Localized>
         <div className={styles.inlineButton}>
-          <FieldText monospace data-name="username">
```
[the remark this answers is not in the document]
nickbrandt: @jimporter this was a change @changecourse made to the designs based on discussions with the Photon team.

src/webextension/locales/en-US/list.ftl
```diff
@@ -3,10 +3,10 @@
 document
   .title = Lockbox Entries
 
-item-fields-title = Site Name
+item-fields-title = Name
 item-fields-title-input
-  .placeholder = Primary Bank
-item-fields-origin = Site Address
+  .placeholder = e.g. primary bank
```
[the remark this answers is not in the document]
nickbrandt: @jimporter this was a suggested placeholder from the eval research, though we can look at other options. @changecourse thoughts?

src/webextension/list/components/item-fields.js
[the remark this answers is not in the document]
jimporter: I disagree with that pretty strongly; there are a number of online accounts I have whose usernames I don't control and which contain a mix of numbers and letters. We use monospace as a way of distinguishing things like 0 (zero) and O (capital oh). This change is a regression for that use case (granted, it's not super-common, but given that the cost is essentially nil, I don't think we should switch to variable-width fonts here).

src/webextension/widgets/dialog-box.js
nickbrandt: @jimporter I've reverted to using `primary` by default. I'll need help in setting `danger` as a prop though.

src/webextension/widgets/password-input.js
```diff
@@ -58,14 +57,10 @@ export default class PasswordInput extends React.Component {
...
         <ButtonStack selectedIndex={selectedIndex}>
-          <Localized id="password-input-show">
-            <button type="button" disabled={disabled}
-                    onClick={() => this.showPassword(true)}>sHOw</button>
-          </Localized>
-          <Localized id="password-input-hide">
-            <button type="button" disabled={disabled}
-                    onClick={() => this.showPassword(false)}>hIDe</button>
-          </Localized>
+          <button className={styles.showBtn} type="button" title="Show" disabled={disabled}
```
devinreams: The "Show" and "Hide" titles are now not localized which means users of other languages will see English. I think we will need @jimporter's help here as I've previously struggled getting params within elements properly localized.

devinreams:
I'm just swinging by to re-state where I think this is at:

- @nickbrandt addressed the majority of @jimporter's review comments
- @jimporter's help is needed on localizing the Show/Hide button (eyeball icon floating on the field)
- @changecourse final decision on removing monospace is required from you. the argument is that by removing it on these fields we will potentially create a poorer reading experience
- @changecourse a final decision on whether the placeholder text for "e.g. primary bank" is good here for now (this is an improvement, but probably still subject to change which is OK by me)
- then a final review (or edits made by) @jimporter and PI would be expected before merging this (for version 0.1.8 release, due next week)

Did I miss(represent) anything? @jimporter to speed things up and cut back on the back/forth are you willing to make some of the final desired changes yourself on this branch?

changecourse:
Monospace should be removed for username field, pending a review with lead designers on Photon and after some back n' forth weighing out pros and cons, in addition to a review of how this field is handled in other areas of Fx, as well as by competitors. 

Placeholder text revision of "e.g. primary bank" was the recommended approach for the entry name placeholder, which came from insights derived by Heather's latest research efforts on alpha.

src/webextension/widgets/password-input.js
jimporter: In the diff for `src/webextension/locales/en-US/widgets.ftl` where localized strings are removed, you can see examples of how to tell Fluent to localize a prop (i.e. something like `title="Show"` in the diff above).

To be nice to future engineers, we (usually) also fake-localize the prop in the JS like `title="sHOw"`; this gets overridden immediately by Fluent, so it doesn't change the behavior, but it's a nice way of documenting what a component's localization should be like. We can also use the fake-localization in unit tests.

jimporter:
> Monospace should be removed for username field, pending a review with lead designers on Photon and after some back n' forth weighing out pros and cons, in addition to a review of how this field is handled in other areas of Fx, as well as by competitors.

Where is this review? Given that the Photon team aren't domain experts in this area, I'm skeptical of their reasoning here, and it would be helpful to see why they think this. I'm especially surprised by this change because we already had a discussion about this and settled on username and password being monospace.

src/webextension/widgets/password-input.js
devinreams: OK thanks for pointer! I will take a quick look at getting this back to a localized state.

devinreams:
>Given that the Photon team aren't domain experts in this area, I'm skeptical of their reasoning here

@jimporter Sure, but they're certainly user interface and design subject matter experts and we respect their expertise just like they do ours. 

I _really_ don't want engineering to prevent interface changes provided by the UR and UX/UI experts here. We can certainly request revisions within reason (code quality, approach, maintainability, best practices).

That said, I know your, mine and other opinions on have been shared, heard, incorporated (and now changed) so I have to trust that the decisions have been litigated, perhaps beyond our sights (and I trust those discussions include the right folks).

I don't know if we should demand much more on this PR without risking wandering into "blocking" territory. This doesn't feel like the kind of thing worth blocking (its all subject to change again, right?).

devinreams:
This branch should be green now ✅  and the localization is back in action! Thanks for the fluent help, @jimporter.

screen shot 2018-03-01 at 12 45 37 pm

Ready for any last UI input on the two concerns above, final code review, and PI review. Nice work @nickbrandt. 👏

m8ttyB:
The size - length - of the password field appears to be a tad bit shorter than the fields above it. The Invision design shows that they should share the same alignment.

screen shot 2018-03-01 at 3 12 58 pm

m8ttyB:
Maybe I'm misunderstanding the scope of the PR, should the `Delete` entry confirmation button be red? I'm seeing it styled as blue.

screen shot 2018-03-01 at 3 26 36 pm

jimporter:
@m8ttyB Yeah, that's something I think I'm on the hook for fixing up.

nickbrandt:
@m8ttyB Thanks for catching that. @jimporter this was an error due to me not updating the styles once 'box-sizing' was removed. Should be good now.

screen shot 2018-03-01 at 5 48 38 pm

jimporter:
@devinreams Ok, let's pull the monospace thing out to a separate issue; I lean towards dropping the change here and making it in the followup, but I can accept it if we make the change and file an issue to potentially revert it.

changecourse:
@devinreams @jimporter can we please accept and close this PR and create a separate issue to consider reverting the monospace change outside of this? The closing of this PR will allow Nick to work on some additional styling issues dependent on these changes. Thanks!

jimporter:
@changecourse This PR awaiting review shouldn't cause any hold-ups on work depending on these changes, especially since the work is mostly-there. It's pretty straightforward to check out a branch based on *this* branch to do work, and then rebase it onto master when this merges. (I do this pretty regularly, and it only starts getting hard to manage when you have a stack of branches 4-5 high.)

The commands you need to do this are just `git checkout -b new-branch` (done from the branch you want to base it on), and then `git rebase -i master` once the base branch has merged.

In any case, once I get back from the DMV, we can discuss the right path to merging this.

devinreams:
>>let's pull the monospace thing out to a separate issue ... I can accept it if we make the change and file an issue to potentially revert it.

Thanks @jimporter 

>can we please accept and close this PR and create a separate issue to consider reverting the monospace change outside of this?

Absolutely, let's continue our process: @jimporter goes through the final code review ("right path to merging this"), make sure its ship-shape 🚢 and get a sign off from PI.

I sent a message in Slack on how to do what Jim described re: continuing work on a separate branch (but based on this one). I'm happy to get on a call or demo in-person anytime today so you aren't blocked on making progress.

changecourse:
Thanks @devinreams... to be explicitly clear, I think we should include the change made in this PR regarding displaying the Username in standard system text, rather than our monospace font, as we previously had it. 

Then, if we wish to file a future issue to capture the concerns around this change, we can do so.

devinreams:
>I think we should include the change made in this PR regarding displaying the Username in standard system text

Understood. To recap from here:

- @jimporter is still "[on the hook](https://github.com/mozilla-lockbox/lockbox-extension/pull/567#issuecomment-369754768)" for the red "danger" button? ⚠️ 
- @jimporter also on the hook for final code review and getting this on "[the right path to merging](https://github.com/mozilla-lockbox/lockbox-extension/pull/567#issuecomment-370003773)" once he returns 🚗 
  - note: @changecourse and @jimporter are good with merging the form/input styles as-is (e.g. the monospace-on-username change, placeholder text from UR) and opening a separate issue or email or whatever to discuss and finalize/document any and all reasoning, sides, opinions, etc. 🤔 
- then PI is up last, but not least, to test, review and incorporate into any other test plans ✅

jimporter:
@devinreams There are a couple other things this needs:

* The notes field is too wide now; I haven't investigated why.
* If I land this with the monospace username removal, then I'd like some assurances that we'll come to a consensus in the near future in a followup issue (i.e. this shouldn't sit in limbo for several sprints).

More generally, I would strongly urge us to work to reach consensus *before* making changes in the future, and I'm only relenting on this particular issue because I don't think we've been especially clear in the past about our process for reaching consensus on changes. I don't want us to get into a habit of unilaterally making a change and merging it before everyone's on board with it (not least because past experience has shown that post-merge fixup/discussion can easily fail to occur).

src/webextension/widgets/input.css
```diff
@@ -28,8 +28,9 @@
 
 .input {
   padding: 0 8px;
-  font-size: 13px;
+  font-size: 15px;
```
jimporter: I know this is what the spec *claims* the font size should be, but the spec is wrong. Per the spec, the inputs should use **Body 20**, which is a `15px` high font. However, if you example the text in the example images in the spec, it's *actually* **Body 10**, which is `13px` high as in our code before this change. In addition, the last I checked, nearly every similar input element in Firefox uses `13px` tall text; since our primary goal in implementing Photon visuals is to look like Firefox, I think `13px` is the better choice.

Let's leave this the way it is for now and raise the issue with the Photon design team. If their intent truly is for inputs to have `15px` tall fonts, I'd like to see that indicated consistently in the specs (or at least an explicit note indicating that their sample images are wrong).

src/webextension/widgets/input.css
```diff
@@ -101,6 +102,27 @@
 }
 
 .label-text {
-  margin: 10px 0 3px 0;
+  margin: 16px 0 4px 0;
   font-size: 13px;
 }
+
+.password {
+  max-width: 258px;
```
jimporter: As above, I don't think this should be here.

src/webextension/widgets/password-input.js
```diff
@@ -59,12 +59,12 @@ export default class PasswordInput extends React.Component {
                ref={(element) => this.inputElement = element} {...props}/>
         <ButtonStack selectedIndex={selectedIndex}>
           <Localized id="password-input-show">
-            <button type="button" disabled={disabled}
-                    onClick={() => this.showPassword(true)}>sHOw</button>
+            <button className={styles.showBtn} type="button" title="sHOw" disabled={disabled}
+              onClick={() => this.showPassword(true)}></button>
```
jimporter: Nit: could you restore the indentation here? `onClick` should be directly below `className`.

src/webextension/widgets/password-input.js
```diff
@@ -59,12 +59,12 @@ export default class PasswordInput extends React.Component {
...
           </Localized>
           <Localized id="password-input-hide">
-            <button type="button" disabled={disabled}
-                    onClick={() => this.showPassword(false)}>hIDe</button>
+            <button className={styles.hideBtn} type="button" title="hIDe" disabled={disabled}
+              onClick={() => this.showPassword(false)}></button>
```
jimporter: Here too.

src/webextension/widgets/button.css
```diff
@@ -71,10 +80,14 @@
...
   height: 32px;
-  min-width: 80px;
   padding: 0 8px;
   font-size: 13px;
   border-radius: 2px;
+  min-width: 80px;
+}
+
+.primary-theme.normal-size {
+  min-width: 132px;
```
jimporter: What's this for? Is there a particular button that needs this? As far as I can tell, it's an optional part of the spec for normal-sized buttons (and doesn't necessarily have anything to do with primary-themed buttons).

I'm not sure *all* normal-sized primary buttons and *no* normal-sized non-primary buttons should have this. I could see adding a sub-theme like `normal-wide` (or maybe even just call it `wide`), but as it is, this seems like the wrong way to do it.

src/webextension/widgets/input.css
```diff
@@ -28,8 +28,9 @@
 
 .input {
   padding: 0 8px;
-  font-size: 13px;
+  font-size: 15px;
   color: #0c0c0d;
+  max-width: 242px;
```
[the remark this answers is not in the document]
jimporter: I see now that the spec intentionally makes the text area wider than the single-line inputs. Setting aside the fact that that looks really broken (I would be unsurprised if we get bugs filed by users about this), the right way to handle this would be to set a `max-width` on all the inputs in `item-fields.js`/`item-fields.css`.

src/webextension/widgets/input.css
nickbrandt: I was just following the spec in this instance (as well as the designs). And I have found instances where 15px font size is used as they define:
image

jimporter: On Linux at least, the exact font size on that field in CSS is `13.3167px`, which gets hinted in height to `14px` (but in width it's about what you'd expect; measuring it, it's about 3% wider than an actual `13px` font, but 10% narrower than `15px`). It may be that there's some platform-specific weirdness happening here too, but based on the inconsistencies in the spec and the OS-level variations we're seeing, I still lean towards implementing the spec as it's *drawn* (i.e. 13px), at least until we get confirmation from the Photon design team about what they really meant in their spec.

I'll file a followup issue and prioritize getting this resolved with the Photon folks, since we really shouldn't have to deal with inconsistent specifications.

nickbrandt: @jimporter Ya that's interesting, on Mac it shows at ~15. I can go ahead and revert this to 13px until we get clarification from the Photon team though.

screen shot 2018-03-05 at 12 07 34 pm

jimporter: Oh, even better. The spec uses multiple different font sizes on the same page for various images showing how the fields should look. >:(

jimporter: At this point, I honestly can't tell what's right or wrong here, or if there should be variations for each OS or what. I think I'm ok with leaving this at 15px, but we really need the spec fixed so we don't have these issues again.

nickbrandt: @jimporter Yea I noticed that as well...wish the styles were defined using code vs. images, as that would make things easier. I'll keep it as 15 then, and Ryan and I can connect with the Photon folks to get this resolved.

jimporter: Filed an issue with Photon here: https://github.com/FirefoxUX/photon/issues/283

jimporter:
Danger button is done. The delete item and reset datastore buttons are dangerous; the cancel editing button is not.

src/webextension/widgets/button.css
nickbrandt: @jimporter I have all of the other changes pretty much wrapped up except for this one. I'm unsure how to setup a sub-theme with your button themes in the button.js file. Is this something you can run through real quick with me either via Slack or Vidyo? Thanks!

jimporter: If you go to `widgets/button.js`, you'll see this block:

https://github.com/mozilla-lockbox/lockbox-extension/blob/956d5bf56299d677b79c21ac2ff21649b090e961/src/webextension/widgets/button.js#L16-L20

You can add an entry like this to it (and then add a rule for `.wide` with your `min-width` in the CSS file):

```
wide: `${styles.normalSize} ${styles.wideSize}`,
```

Then you just have to find the buttons you want to be wide in the appropriate JS files and add `theme="wide"` to them.

nickbrandt: Gotcha, thanks for the help!

nickbrandt: This all seemed good, except you meant to use `size="wide"` instead of `theme="wide"`, correct? As long as that is right I'll get this committed and ready for review.

jimporter: Right, sorry; `size="wide"` is what you want.

src/webextension/list/components/item-fields.css
```diff
@@ -17,6 +17,18 @@
   resize: none;
 }
 
+.item-fields input {
+  max-width: 242px;
+}
+
+.item-fields .password {
```
jimporter: The `.item-fields` part here isn't actually necessary, since `.password` uniquely identifies the element you want (class names get decorated so that there aren't collisions from definitions in other files).

src/webextension/list/components/item-fields.css
```diff
@@ -17,6 +17,18 @@
   resize: none;
 }
 
+.item-fields input {
+  max-width: 242px;
```
jimporter: Could we change this to `.item-fields .input` (or even better, simply `.input` like above)? Then you'd just need to pass `className` to all the inputs in `item-fields.js`.

I know this is a really minor thing, but technically, the `input` element is contained within a different React component, and it's not actually a part of that component's public API (in other words, we might change the implementation of that component one day in a way that breaks this). However, passing `className` to override some of the default styles for that component *is* part of the public API, and we make a guarantee that that should always work.

nickbrandt:
@jimporter Got it, thanks for pointing those out to me!

changecourse:
@devinreams Is there a reason why this didn't get pulled into the done column once we closed it?

devinreams:
>Is there a reason why this didn't get pulled into the done column once we closed it?

@changecourse I see it there now. The removal of “in progress” label should’ve been the mechanism to do that but sometimes API events get missed. I will monitor this kind of stuff as part of release management...

Issue #497, opened 2018-01-25
PI - Scope accessibility test plan requirements


m8ttyB:
Scope accessibility testing into individual stories that represent the various layers of inclusiveness.

The accessibility test plan will follow these guidelines https://docs.google.com/document/d/145Ce8SDqVmVWqe0RPGTnF-fxvZjwnop9EiM9L6WzVyw/edit?usp=sharing

m8ttyB:
@kimberlythegeek check-pointed with me today, the scoping has begun, once drafted she'll attache a link to a Google Doc.

kimberlythegeek:
Here's my current draft of the test plan. Also the [google doc is here](https://docs.google.com/document/d/1P71BtThg9rC9kXdGen9JRx46_QnTx033WLNZ4F1DEMU/edit?usp=sharing).


# Manual Test Plan

## KEYBOARD NAVIGATION

Navigate the web content using only the keyboard
- Tab Order: Verify that `TAB`, as well as `SHIFT + TAB`, follows a logical and intuitive order.
- Do all controls, such as links, buttons, etc., get focused? _(Focus should be visibly apparent)_
- If a modal or pop-up window opens does focus shifts to the pop up? 
  - Is focused constrained within the modal?
  - Can the modal be exited via keyboard? (`ESC` or cancel/exit button)
  - When a pop up window is closed, does focus return to a logical point? _(Such as where focus left off)_
- Does the webpage contain a "skip to content" link at the top of page? _(This may or may not be visible)_
- Ensure that all content that is visually hidden, and not meant to be available (especially in dynamic web applications) is also hidden from the keyboard and/or a screen reader.
  - Using `hidden` or `aria-hidden='true'` if necessary. _(See [hidden and aria-hidden](https://developer.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/))_

## PAGE STRUCTURE
- Do hyperlinks have text that give context? _(Can the purpose of the be understood in an isolated context? As opposed to a link reading "click here”)_
- Do all images contains an alt description? _Such a description should concisely explain the purpose of the image to a non-sighted user._
- Has a logical hierarchy of header tags been used?
- Does the page utilize [ARIA landmark roles](https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/aria-landmarks/)?
- Have ARIA attributes been put into use?
[...]

PR #572, opened 2018-02-23
Add a11y test plan to docs

kimberlythegeek:
Added test plan from #497 to docs

### DIFF
--- a/docs/accessibility-test-plan.md
+++ b/docs/accessibility-test-plan.md
@@ -1,0 +1,48 @@
+# Manual Test Plan
+
+## Keyboard Navigation
+_< 3 hours_
+
+Navigate the web content using only the keyboard
+- Tab Order:`TAB`, as well as `SHIFT + TAB`, follows a logical and intuitive order.
+- All controls, links, buttons, etc., get focused.
+- Focus should be visibly apparent.
+- When a modal or pop up window opens, focus shifts to the pop up.
+  - Focus and tab order is constrained within the modal.
+  - Modal can be exited via keyboard.
+  - When a pop up window is closed, focus returns to a logical point.
+- Ensure that all content that is visually hidden is also hidden from the keyboard and/or a screen reader. _Except content specifically for screen reader users._
+- Firefox toolbar icon can be accessed by keyboard and screen reader users.
+
+## Page Structure
+_3 hours_
+- Links and buttons have text that give context.
+- All images contain alt text to describe the purpose or content of the image to a non-sighted user.
+- A logical hierarchy of header tags has been used.
+- [ARIA landmark roles](https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/aria-landmarks/) have been utilized where applicable.
+- ARIA attributes have been put into use.
+  - Any custom controls have been given the proper role attribute.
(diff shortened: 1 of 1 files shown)


test/integration/pages/home.py
```diff
@@ -26,7 +26,7 @@ class Home(Base):
     _new_entry_locator = (By.CSS_SELECTOR,
                           'section menu '
                           'button.{}:nth-child(1)'.format(
-                            munged_class_name('button')))
+                              munged_class_name('button')))
```
jrbenny35: Any reason for this change?

jrbenny35:
There are a lot of commits for a simple doc addition, any way you can reduce them?

docs/accessibility-test-plan.md
```diff
@@ -0,0 +1,48 @@
+# Manual Test Plan
+
+## KEYBOARD NAVIGATION
```
jimporter: Editorial nit: the all-caps on these heading isn't consistent with the rest of our docs. We generally use Title Case (or Sentence case; we should pick one and stick to it).

test/integration/pages/home.py
jimporter: When looking at this and the changes below, the new indentations seem more consistent, so I'm happy with it.

docs/accessibility-test-plan.md
```diff
@@ -0,0 +1,48 @@
...
+## UX Design Considerations
+- Interactions available with a mouse are also available using a keyboard.
+- Information conveyed visually is also available by other means.
+  - Pop-ups and other state changes are announced to screen readers using appropriate ARIA attributes.
+  - Temporary on-screen events give users ample time to receive conveyed information.
+- Animations are not excessively flashy, as this can cause seizures in some users.
+
+## TOOLS AND RESOURCES
+Use a variety of testing tools to analyze pages:
+- aXe developer tools by Deque
```
pdehaan: Is this just the Firefox/Chrome add-on, a la https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/, or were you using some other aXe tools?

docs/accessibility-test-plan.md
```diff
@@ -0,0 +1,48 @@
...
+- Browser developer tools to inspect HTML for logical heading structure and ARIA roles
+- [Color blind simulators](https://www.toptal.com/designers/colorfilter)
```
pdehaan: Awesome resource, thanks!

pdehaan:
> There are a lot of commits for a simple doc addition, any way you can reduce them?

When we merge this, somebody can press the "Squash and merge" button, which should reduce this down to 1 commit (without needing to do a messy rebase).

pdehaan:
LGTM. I agree with @jimporter's comment about converting the H2 headings to use sentence case rather than all uppercase, for consistency.

test/integration/pages/home.py
kimberlythegeek: I didn't realize I included these changes. I was only trying to push the new documentation file.

I got under-indentation errors when running flake8, which is the reason for these edits.

docs/accessibility-test-plan.md
kimberlythegeek: There are the browser extensions, but also the axe-core API, which is used by those extensions.

I've been working on a python package that uses axe-core to run automated accessibility tests.

PR #595, opened 2018-02-28
Set autoFocus on search filter input so it's focused immediately

### LINKED ISSUE #573: "Search Lockbox Entries" field should auto-focus
The "Search Lockbox Entries" fields in the Lockbox door hanger and "Manage Lockbox" page should auto-focus when they open. It is a frustrating break in my workflow to have to manually mouse over and click to focus the "Search Lockbox Entries" field instead of just immediately typing after opening the door hanger or "Manage Lockbox" page.

devinreams:
Resolves #573

I realize this is neither prioritized work nor are we agreed on what/how to address this, but having to tab or mouse to the search field 99% of the time I'm opening Lockbox (doorhanger _or_ manage list) really slows _my_ day-to-day usage.

Here's a one-line change to autofocus on the search field both in the doorhanger and the manage listing.

autofocus

I figured figured a one line change wouldn't be _too_ problematic. I realize this could be abstracted into a property (so we can explicitly set it per use) but when looking at the latest designs, the search box is going to be the first element in the top right of the interface in both manage list and doorhanger so figure we just go ahead and drop the user in it without abstracting stuff too much (aka wait until we need it).

### DIFF
--- a/src/webextension/widgets/filter-input.js
+++ b/src/webextension/widgets/filter-input.js
@@ -55,7 +55,8 @@ export default class FilterInput extends React.Component {
       <div className={finalClassName}>
         <input type="search" {...props} disabled={disabled}
                value={this.state.value}
-               onChange={(e) => this.updateValue(e.target.value)}/>
+               onChange={(e) => this.updateValue(e.target.value)}
+               autoFocus/>
         <Localized id="filter-input-clear">
           <button type="button" disabled={disabled}
                   onClick={() => this.updateValue("")}>


devinreams:
>Unfortunately, this is a pretty big hack and I don't think it's something we could land.

OK, I was afraid of that. I'll close this PR since it's not going to get us any real value and keep the original issue open #573 for future prioritization. I figured if it was easy, just do it, if not.. well, here we are.

PR #596, opened 2018-02-28
Include whitespace in item fields so new lines are output

### LINKED ISSUE #504: Newlines in Entry Details Notes are not displayed when viewing an entry.
I added some Notes to an Entry Detail using newlines to separate some key/value pairs. In "Edit Entry Details" mode, they look something like:
```
Fruit: Banana
Dessert: Ice Cream
```
But in the view mode, the newlines are not displayed, e.g.:
```
Fruit: Banana Dessert: Ice Cream
```

devinreams:
Resolves #504 

newlines

If this is too heavy-handed and the `white-space` style should be scoped differently, I'm open to that. 

I also confirmed the semi-related-but-not-really issue with [overflow on really long items](https://github.com/mozilla-lockbox/lockbox-extension/pull/409#issuecomment-351463521) that @m8ttyB found a couple times does not revert here. 🤞


jimporter:
Ok, another option, which I kind of like and am leaning towards a bit, is for us to set this CSS rule on our `<FieldText/>` component. That component is there to display user-entered text from a form, and I think it makes sense for us to show spaces/newlines exactly as the user entered them. This would also make it possible for users to see if they have, say, multiple consecutive spaces in their password (I'm not sure *why* anyone would do this, but hey).

devinreams:
>Ok, another option, which I kind of like and am leaning towards a bit, is for us to set this CSS rule on our <FieldText/> component. 

@jimporter Yeah, that makes even better sense. Thanks for the quick feedback.

I _think_ that means I just move this to `widgets/input.css` on "input-text"? I commited now but if I'm being dense please just let me know where to put it and we can wrap this up. 😬  

A quick re-test of the output:

screen shot 2018-02-28 at 11 41 17 am

jimporter:
> I think that means I just move this to widgets/input.css on "input-text"?

Not `.input-text`, `.field-text` (you got it right in the code :)). But otherwise, yes.

devinreams:
Thanks @jimporter.

Waiting to merge on final PI review and approval..

PR #597, opened 2018-02-28
Styling updates based off Jim's code review of #549

nickbrandt:
-Updated pseudo-elements to have two colons instead of one
-Moved toolbar margin update to be specific to navigation bar
-Removed arrowhead down icon that is no longer in use
-Updated dropdown hover, focus and active states after reviewing how Firefox browser handles similar items

image


src/webextension/widgets/toolbar.css
```diff
@@ -10,7 +10,7 @@
 }
 
 .toolbar > * + * {
-  margin-inline-start: 5ch;
+  margin-inline-start: 20px;
```
jimporter: With this change, could you also update the padding on the `.toolbar` to be `10px`? That way, we're using consistent units.

devinreams:
@nickbrandt Do you think your changes warrant PI review?

If, so adding the team as a "Reviewer" here and adding the "needs-pi" label appears to be needed to indicate that (still catching up, correct me if wrong!).

PR #600, opened 2018-02-28
Automatically focus the filter input when the page loads; resolves #573

### LINKED ISSUE #573: "Search Lockbox Entries" field should auto-focus

jimporter:
This PR auto-focuses the filter input boxes, but *only* when the entire page loads (using `autoFocus` will trigger a focus change any time the component is mounted, which has some bad UX on the doorhanger).

### DIFF
--- a/src/webextension/list/containers/item-filter.js
+++ b/src/webextension/list/containers/item-filter.js
@@ -3,19 +3,24 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { Localized } from "fluent-react";
+import PropTypes from "prop-types";
 import React from "react";
 import { connect } from "react-redux";
 
 import { filterItems } from "../actions";
 import FilterInput from "../../widgets/filter-input";
 
-function ItemFilter(props) {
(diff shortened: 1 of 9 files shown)


jimporter:
> This looks fine, but I am starting to wonder if we have a misunderstanding of how autoFocus works in React now.

Entirely possible, since there's no documentation in React for `autoFocus` at all. :/ I'm guessing it just delegates to the browser, which in practice appears to mean "whenever the parent component is mounted, this thing will be focused", which isn't always what we want.

jimporter:
Force-merged after rebase.

jimporter:
Oh wait, darn. I thought this one had PI review already... sorry about that, @m8ttyB! If you want me to back it out, just let me know.

jimporter:
And we all lived happily ever after!

PR #564, opened 2018-02-22
Doorhanger sign-in styling adjustments

nickbrandt:
Resolves #482 
Updated doorhanger sign-in styles to reflect design
Added new button theme and size to account for these buttons as well as others in designs

PR #601, opened 2018-03-02
Create a <PanelFooterButton/> and update the styles for the unlock do…

### LINKED ISSUE #482: DR: Align doorhanger sign-in to Photon spec
*The doorhanger you see if you are a returning LB user, but are not authenticated*

**Inivision: https://mozilla.invisionapp.com/share/W3EU997YN#/264182186_D_1-1-0_Default_-_Locked_State**

## Tasks ##
- [x] Match to spec provided in above design (padding, sizing, layout)

jimporter:
…orhanger; resolves #482 

@nickbrandt @changecourse This is an alternate version of #564; it refactors the unlock doorhanger a bit to use the Panel widgets created for the list-view doorhanger. This should help us keep them looking thematically-similar. I made a couple of visual changes as well:

* The panel buttons have 0 border radius, since the doorhanger has square borders. Trying to apply a radius to the buttons looked subtly wrong.
* I adjusted the focus state for the buttons to only have an inner-border so they don't look broken. However, this may be a bit too subtle for the primary button.
* I picked an arbitrary value for the border on the footer; it's a bit darker than the spec, but it helps us use the same color as an overlay for darkening the edges of the primary button. We could change this, though.
* I applied the same border from the footer to the panel header as well.
* I made the button on the list-view doorhanger into a panel footer button, so it uses the new styles and generally looks nicer.

I still need to write a test for `<PanelFooterButton/>`...


src/webextension/unlock/components/app.css
```diff
@@ -5,70 +5,40 @@
 body {
   padding: 0;
   margin: 0;
-  width: 344px;
+  width: 320px;
   color: #0c0c0d;
   background-color: #f9f9fa;
```
devinreams: @jimporter I think this needs to change to`#ededf0` from `#f9f9fa`?

firefox_nightly_and_d_1_1_0_default_-_locked_state

src/webextension/unlock/components/app.css
```diff
@@ -5,70 +5,40 @@
...
   font-size: 13px;
   -moz-user-select: none;
 }
 
-.unlock {
-  margin: 0;
-  display: flex;
-  flex-flow: column nowrap;
-  align-items: center;
+.lockie {
```
devinreams: Nice 🐰🥚

jimporter:
> We also appear to be missing the waves and ripples from the svg? (as seen between the two screenshots)

We'll need a new SVG then, since this one uses the same color for the line-art as the background.

src/webextension/unlock/components/app.css
jimporter: Fixed. (I wasn't testing with the real doorhanger; I usually just open up the page directly in the browser, since it's faster to get to.)

devinreams:
>We'll need a new SVG then

@changecourse @nickbrandt would you be willing and able to add this (and any other relevant / recent) assets to https://github.com/mozilla-lockbox/assets so we can easily get to those and self serve on this kind of stuff? let me know if you'd like an extra hand and I can help get stuff loaded in there.

nickbrandt:
@devinreams @jimporter Updated Nessie SVG has been added to the assets folder (https://github.com/mozilla-lockbox/assets/blob/master/images/nessie.svg). Going forward will start making sure updated assets also make it here.

devinreams:
>This looks good to me once the Nessie SVG is updated!

@nickbrandt @changecourse Ah cool. It looks like this version has the "waves" color go from `#eaeaea` but the comp shows something closer to `#d7d7db`.

Right now we use one SVG for all three instances of Nessie (getting started first run, "home page" placeholder, and the doorhanger). Should this be universally replaced or just for this doorhanger instance?

changecourse:
@devinreams it should be replaced universally.

devinreams:
>it should be replaced universally.

Here's the latest Nessie on the doorhanger 🐳 

screen shot 2018-03-02 at 1 07 47 pm

@changecourse @nickbrandt just to double check: did you have any final thoughts based on what @jimporter posted in the original description of this PR?

Either way, a final 👍 from either of you would be appreciated and I'll add PI for their review here, too. 😁

changecourse:
Darker line is fine, approved 👍

devinreams:
This is green all over and ready to merge! Thanks, everyone. ✅ 

I'll let @jimporter do the honors. :shipit:

PR #603, opened 2018-03-02
Make sure options_ui is fully displayed on about:addons

### LINKED ISSUE #569: Reset Lockbox button displays as cut off
When looking at Lockbox Preferences, the 'Reset Lockbox' button displays as cut off even when browser window is maximized. There is a hidden scrollbar that allows the user to see the full button, but it shouldn't be needed when browser window is at maximum width/height.

reset

devinreams:
Fixes #569 

add-ons_manager_and_add-ons_manager


src/webextension/settings/components/app.css
```diff
@@ -2,6 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+html {
+  /* make sure this fully displays on the about:addons view */
+ flex-direction: column;
```
jimporter: I don't think this should be necessary; there's only one child element anyway.

devinreams: >there's only one child element anyway.

Good point. I think I was playing around with some different attempts and kept this unnecessarily.

devinreams:
>This doesn't entirely resolve the issue. There are two remaining problems...

Good catches. I admit I didn't explore this enough. Thanks for the extra eyes.

Latest commits are ready for a final review...

devinreams:
I fixed the merge conflict from master and took the `margin: 1em;` @jimporter recently introduced on master (versus my `5px`).

I'm merging with prior approval and green tests. ✅

jimporter:
Ha, that change might have accidentally slipped to one of my PRs in when I was testing out my suggestions for this review. Who can remember anymore? :/

PR #607, opened 2018-03-02
Updating styling of 0 entries page to reflect designs

### LINKED ISSUE #478: DR: Update the '0 entries' promo screen to align to photon
*Adjustments to the first-time use, 0-entry screen*

**Inivsion: https://mozilla.invisionapp.com/share/W3EU997YN#/264245494_1-1-0_Lockbox_Home_-_0_Entries**

## Tasks ##
- [x] Change shadows on promo images to ‘Shadow 10’ instead of 30 in spec
- [x] Line-spacing for promo descriptions should be 22.5 (1.5x)

nickbrandt:
Resolves #478 

-Intro image shadow to Shadow 10
-Updated image border
-Defined line height for intro paragraphs

### DIFF
--- a/src/webextension/list/manage/components/intro-page.css
+++ b/src/webextension/list/manage/components/intro-page.css
@@ -8,7 +8,7 @@
   grid-template-columns: 1fr 1fr 1fr;
   grid-template-rows: auto auto auto;
   grid-column-gap: 4em;
-  margin: 4em auto;
+  margin: 8em auto 4em;
   padding: 0 4em;
   max-width: 800px;
 }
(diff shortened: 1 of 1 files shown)


nickbrandt:
@jimporter Can you clarify on your comment regarding line height? Most all of the design actually uses 1.5 line height for paragraphs, headings, etc. I would argue this should be more of a global change, but noticed you have been doing these sort of declarations specific to elements in other areas of the app. What would you prefer?

jimporter:
@nickbrandt Assuming it's supposed to be a global thing, I think we should probably just specify the line-height in the `body` for all our documents. I don't really recall why we specified line-height individually in all those CSS rules (looking at the blame, it's probably something I missed in the review back then). However, that's probably a bigger change and we should do it in a separate cleanup issue so PI can verify that we have identical before/after visuals, so feel free to ignore my previous concerns. :)

devinreams:
>>I noticed you have been doing these sort of declarations specific to elements in other areas of the app

I know _I_ did this per-element since I wasn't confident enough in the "global" styles we had at the time. 

>I think we should probably just specify the line-height in the body for all our documents

Good call. Follow-up issue created here: https://github.com/mozilla-lockbox/lockbox-extension/issues/614

Issue #521, opened 2018-02-05
Implement URL parser


sandysage:
*Based on https://gist.github.com/linuxwolf/5722b60907826b47a97b10cf36676ca2
*Use Mozilla public suffix list: https://publicsuffix.org/

sandysage:
This needs to come before 'fill'.

jimporter:
Structurally, my plan here is to implement a `matchURL()` function that takes a query and an entry, and returns a bool indicating whether it matched. Then we can apply that function over the list of items to determine the matched items. This way, a given query can be mapped to multiple alternatives (e.g. `wwww.facebook.com` and `facebook.com`) as well as doing "fancy" things with checking port numbers and such.

One possible enhancement would be for `matchURL()` to return an integer score ranking the match. That way, better matches are higher-ranked and thus listed earlier in the results. That's something I think we could delay for a while, since it should only be noticeable in certain edge cases.

linuxwolf:
@jimporter some questions/comments:

1. What is the query?  Is this the full URL as retrieved from the active tab (for instance), or is it something shorter?
2. While a stretch goal, what are the edge cases you see would be mitigated by a scoring system?
3. We need to adhere to origin as much as possible, which includes scheme and port.  Not doing so can (and has) led to vulnerabilities in other systems.

jimporter:
@linuxwolf:

1. The query would be everything before the path in the URL (minus HTTP basic auth username/password, of course), so basically `URL.protocol + "//" + URL.host`; that will give us protocol, hostname, and port (with port being implicitly-defined if it's the protocol's default). That should be sufficient for any searches we want to do, while also looking readable and being easy enough for a user to type on their own.
   1. Our `matchURL()` function would accept a wider variety of queries (i.e. user-provided ones), but would effectively transform them into a query of the above format before doing anything. For example, if a user enters `www.facebook.com`, it will be as if they entered `https://www.facebook.com` (which in turn is equivalent to `https://www.facebook.com:443`).
2. Scoring could help if you have accounts on subdomains that you want to be higher-priority than an account on the root domain. I can't think of many real-world sites that do this, but it may help for some intranet sites.
[...]

Issue #515, opened 2018-02-02
Provide a full list of options when there are no-results for the current URI to mitigate confusion for why there aren't relevant results applicable to that URI


sandysage:
We'll have to add a distinction between user-entered filters and the autogenerated filter from (1).

This will probably be something like a banner that shows up below the search box that says "no results", with the full list of entries below

devinreams:
Closed at #646

PR #612, opened 2018-03-05
Prepare 0.1.8 release

devinreams:
Work in progress to start preparing the release and notes.

- [x] bump version
- [x] create 0.1.8 release notes section
- [x] update release notes based on final code
- [x] review from product on notes verbiage


docs/release-notes.md
```diff
@@ -1,15 +1,44 @@
...
-_Dtae: 2018-02-23_
+_Date: 2018-02-23_
 
 This release is to address a critical defect from the previous release.
 
 ### What's Fixed
 
 * If you linked your Firefox Account, Lockbox would stop working when you restart Firefox ([#568](https://github.com/mozilla-lockbox/lockbox-extension/issues/568))
 
+### Known Issues
```
linuxwolf: thank you (-:

devinreams:
@sandysage we can take one final pass after sprint review this morning but I believe this captures everything (scheduled) to land in 0.1.8:

https://github.com/mozilla-lockbox/lockbox-extension/blob/prepare-0.1.8/docs/release-notes.md

Can you make any edits you'd like and give your final 👍 approval when ready?

docs/release-notes.md
```diff
@@ -1,15 +1,46 @@
...
 
+## 0.1.8-alpha
+
+_Date: 2018-03-08_
+
+### What's New
+
+- Added support For Telemetry Scalars so we can easily tell how many items are stored in your Lockbox datastore ([#552](https://github.com/mozilla-lockbox/lockbox-extension/pull/552))
+- Made it so the search box is automatically focused so you can start typing to search your entries immediately ([#573](https://github.com/mozilla-lockbox/lockbox-extension/issues/573))
+- The doorhanger also automatically fills the search with the domain of the current page you're viewing to help you quickly see matching entries ([#616](https://github.com/mozilla-lockbox/lockbox-extension/pull/616))
```
m8ttyB: \o/

devinreams: @sandysage would you like to acknowledge the "limitations" of the autofill here? or in the known issues section? either way I didn't add anything and leave it to your judgement on what/how we should disclose. 😀

docs/release-notes.md
```diff
@@ -1,15 +1,46 @@
...
+- Improved interface colors and styles for entry list items ([#553](https://github.com/mozilla-lockbox/lockbox-extension/pull/553))
+- Improved search bar interface style with a simpler "Clear" button ([#562](https://github.com/mozilla-lockbox/lockbox-extension/pull/562))
+- Re-ordered the create/edit entry interface with improved spacing and styles ([#567](https://github.com/mozilla-lockbox/lockbox-extension/pull/567))
+
+### What's Fixed
+
+- Item notes with line breaks properly show those separate lines when editing ([#596](https://github.com/mozilla-lockbox/lockbox-extension/pull/596))
+- The "reset Lockbox" settings interface fits better in about:addons ([#603](https://github.com/mozilla-lockbox/lockbox-extension/pull/603))
+- Fixed alignment on the entry detail view so the buttons don't jump when switching between edit and view modes ([#608](https://github.com/mozilla-lockbox/lockbox-extension/pull/608))
+- Updated development and testing dependencies via Greenkeeper.io and pyup.io: documentation 6.0.0, selenium 3.10.0, flake8-docstrings 1.3.0, pytest-xdist 1.22.2, pypom 1.3.0, pytest-selenium 1.11.4, pytest 3.4.1, flake8-isort 2.4
```
linuxwolf: Previous release entries listed dependency updates in a sub-list.  In hindsight they could have been more concise on the wording, but breaking them out makes it clearer what was updated.

devinreams: I hesitated even including this at all. The question I ask myself: do we really want or need to provide insight into dependencies in a user-facing document? I'd argue no, it provides no value I can think of.

@sandysage I assume you had thought about this before. Would you like to strike this entirely? Or keep it and make it consistent with what we've done in the past? (or Option C, something else?)

docs/release-notes.md
sandysage: @devinreams Yes!

The doorhanger automatically fills the search with the domain, which allows for future improvements to accommodate when there are no results (#515) and parsing the results (#521)...stay tuned.

docs/release-notes.md
sandysage: If the dependencies aren't providing a user value, at least at this time, let's strike them. Is there ever an instance where these dependency updates do provide users value? Eg your password manager is always up to date to avoid security vulnerabilities.

linuxwolf: The user value is not obvious ... probably entirely in the realm of problems dodged in the first place by incorporating defect fixes and other improvements before they are noticed by users.

devinreams: I don't think there's a good argument for these dependencies to be included this time then (mostly python stuff for integration tests, and documentation package for our docs site.. not even this extension). I'll strike it for now, but we can always add it back.

I'll continue to highlight these in the future so we can decide if there's something worth sharing with our users in these notes...

docs/release-notes.md
sandysage: Let's just leave this as is. Everything else checks out and it's not worth the changes here. Besides, my note might just be opening a can of worms unnecessarily.

devinreams: >Let's just leave this as is. E

@sandysage Hah whoops I didn't see this until now... I think we're done here! You can always go back and edit if you'd like. 😁

Issue #520, opened 2018-02-05
Provide a quick method of copy/paste when there are one or two results for a URI to help users log into their account quickly


sandysage:
No distinction needed at this stage for whether a user types in a search result versus we determine it based on the URI.

Goals: provide a way for people to more-quickly copy the username/password.

jimporter:
@linuxwolf I've been looking at this; one thing that makes this somewhat difficult is that so far, I've attempted to minimize the cases where we hold passwords unencrypted in memory. Currently, this means that we only fetch the *full* item details (including the password) when we load the details view for that item. However, this issue proposes giving users access to the password *before* the details view is opened. The simple solution would be pre-fetching all the passwords, but I'm not entirely comfortable with that.

One option would be to change how our copy buttons work so that, instead of simply passing them a string that they put into the clipboard, it actually initiates a fetch of the full item from the datastore and only *then* puts the data in the clipboard. (We could still do the simple "synchronous copy" when we already have the relevant data available though.)

Do you have any opinions here?

linuxwolf:
@jimporter We should change out copy works; fetch the relevant item, populate the clipboard, forget the item.  The only time we should have the data already cached is from the detailed view in the first place.

I don't think we should be concerned about potential latency here, at least not yet.

jimporter:
@linuxwolf Ok, that's what I was thinking. We could even go a step farther and only have the password in memory when the details view is open *and* the user clicked Show Password, but that's probably a separate issue.

linuxwolf:
I think we can start with documenting it (on the list for the threatened threat-model doc from daily standups), and decide later how much of an issue it really is.

jimporter:
Hm, how did this not get auto-resolved?

devinreams:
>how did this not get auto-resolved?

@jimporter I believe it's because the [auto-close keyword feature](https://help.github.com/articles/closing-issues-using-keywords/) only "connects" on commit messages and PR descriptions, but not PR titles 😔

PR #616, opened 2018-03-06
Prefill the URL in the doorhanger

### LINKED ISSUE #514: Get the URL to the doorhanger and fill it into the search box to provide relevant results for the current URI
Per the notes in #498

At this point, we can just take the hostname (no protocol, no port) for simplicity.

This will probably be handled by pulling the data (i.e. when the doorhanger opens, we'll use WebExtension APIs to grab the active tab's URL and then rip out the bit we care about).

Child of #317

jimporter:
Resolves #514 

I think this is close to done for an initial implementation of the URL prefilling. Perhaps I should move the stuff in `list/popup/index.js` to a separate file so I can write a unit test for it though...

### DIFF
--- a/src/webextension/list/containers/item-filter.js
+++ b/src/webextension/list/containers/item-filter.js
@@ -24,7 +24,7 @@ ItemFilter.propTypes = {
 
 export default connect(
   (state) => ({
-    value: state.filter,
+    value: state.list.filter,
     disabled: state.cache.items.length === 0,
   }),
   (dispatch) => ({
--- a/src/webextension/list/manage/components/app.js
+++ b/src/webextension/list/manage/components/app.js
@@ -21,7 +21,7 @@ import styles from "./app.css";
 
 export default class App extends React.Component {
   componentDidMount() {
-    this._filterField.focus();
+    this._filterField.focus(true);
   }
 
   render() {
--- a/src/webextension/list/popup/components/app.js
+++ b/src/webextension/list/popup/components/app.js
@@ -12,7 +12,7 @@ import styles from "./app.css";
 
 export default class App extends React.Component {
   componentDidMount() {
-    this._filterField.focus();
+    this._filterField.focus(true);
   }
 
   render() {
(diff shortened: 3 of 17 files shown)


jimporter:
> Are we okay with not stripping off the `www` in url before putting it into the doorhanger search filter?

Yes. We discussed this a few weeks ago, and improving the URL extraction is #521, which is part 4 of 4 for the URL autofill project. (This is part 1, and I'm starting work today on part 2, which is #520.)

m8ttyB:
Thx @jimporter, yeah this feels like product call. Apologies for not understanding the known limitations, too many child tickets, I missed the nuance of this being broken out across several tickets.

m8ttyB:
Just want to verify one final set of behaviors: if the URI matches an entries site name, site url, or username the door hanger will filter on those entries. Is this this expected per the designs?

screen shot 2018-03-07 at 8 20 34 am

linuxwolf:
@m8ttyB yes, that behavior is expected with this iteration.

This is the first of a number of `task`s that cover the progression to the ideal behavior, as has been covered in planning before.

devinreams:
>There is, however, an argument to be made that this PR should land at the beginning of a release cycle so we have time to fix some of the UX badness

devinreams:
>>There is, however, an argument to be made that this PR should land at the beginning of a release cycle so we have time to fix some of the UX badness

Yup, that's fair. I selfishly wanted it in this release to help speed up my actual day-to-day use of the extension.

@sandysage it's up to you: do you think the "cost" of potential confusion to users outweighs the benefit this initial/progressive feature would offer (convenience when retrieving)?

sandysage:
> do you think the "cost" of potential confusion to users outweighs the benefit this initial/progressive feature would offer (convenience when retrieving)?

As a general rule (that I'm sure we'll break on occasion) let's continue to keeping things moving forward quickly; keeping to our release cycles. We'll capture the WIP state in the release notes, and see how much screaming actually happens :).

jimporter:
Rebased and merged; the only code change was an additional comment as suggested by @linuxwolf

PR #619, opened 2018-03-06
DH Entry List Styling Adjustments

### LINKED ISSUE #483: DR: Doorhanger entries list, align to photon
*Many of the changes requested in this story may be inhereted by default from the changes requested to the entries list of the editor view*

**Invision: https://mozilla.invisionapp.com/share/W3EU997YN#/262323598_D_1-1-0_Default_-_Entries_List**

## Tasks ##
- [x] Expand padding of search box within doorhanger
- [x] ~~Change placeholder content to remove “Entries”~~ -included with #562 
- [x] ~~Styling of doorhanger footer “Manage Lockbox” to match designs (full row clickable)~~ included with #601 
- [x] Decide on max/min-width for default doorhanger experience (what is implemented is narrower than what was posed in designs) 320px is posed --Updated to 320 to match sign-in DH
- [x] Inherit the updates to entry cards from full-tab revisions (height, chevron, etc)
- [x] Focus state of an entry within the doorhanger list (I couldn’t activate it via keyboard)

nickbrandt:
Resolves #483 

-Added padding to DH filter bar
-Updated width of DH to match designs and sign-in sizing
-Wrapped DH filter in <Toolbar/> in preparation for add button as in the full view
-Moved styles related to entries to apply to both DH and full view
-Updated Panel BG color


src/webextension/list/popup/components/app.css
```diff
@@ -13,6 +13,6 @@ body {
 }
 
 .app {
-  width: 250px;
+  width: 320px;
```
jimporter: This is probably something for @changecourse to think on, but I'm of two minds about this:

1. It's good that this is consistent with the Unlock doorhanger
2. This might be too wide for a usable interface; if the doorhanger is so wide that it covers up the login form on a page, it reduces the usability of the doorhanger because now you have do the following:
    1. Open the doorhanger
    2. Copy the password (at least 1 click when we have URL prefilling, but at least 2 right now)
    3. Close the doorhanger
    4. Right-click on the field
    5. Select Paste
   ---
    With a narrower doorhanger, you can merge (iii) and (iv) together, reducing the number of clicks by 1. On a 720p monitor, 320px is *just enough* to force users to take the longer path to pasting their password into Facebook. Here's an example of the issue. You have about 1 or 2 pixels where you can actually click on the password field while the doorhanger is open; if we reduced this by even 20 pixels, it would be a significant usability improvement for a very popular website.

    too-wide

src/webextension/list/popup/components/item-list-panel.js
```diff
@@ -23,8 +26,10 @@ export default function ItemListPanel({inputRef}) {
 
   return (
     <Panel>
-      <PanelHeader>
-        <ItemFilter inputRef={inputRef}/>
+      <PanelHeader className={styles.searchHeader}>
+        <Toolbar className={styles.filterBar}>
```
jimporter: I'd say that the `<PanelHeader/>` should *implicitly* work like a toolbar, so let's remove this line and modify the styling for the `<PanelHeader/>` to be more toolbar-y, if that's a word.

src/webextension/list/popup/components/item-list-panel.css
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+.filter-bar {
+  padding: 12px;
+}
+
+.search-header {
+  height: auto;
```
jimporter: I'm not a big fan of this because it means that the header on the list panel and the header on the details panel are different heights. It'd be great if we could get them to be the same height. This is another thing for @changecourse to think about, but I've been considering how we'd handle transitions between the list and details view of the doorhanger.

From my experience with mobile UI design on Firefox OS, the doorhanger looks remarkably similar (the various Panel components I designed are very much meant to mirror Firefox OS and other mobile layouts). With that in mind, I'd *like* it if the headers were the same height (*and* the same color) so that it appears that these are actually the same element. For transitions, we could then, say, immediately replace the header (so it appears that just the text was updated) and then slide in the detail panel from the right (using the Photon animation curve, of course). We'd do the reverse animation when hitting the Back arrow.

This does, however, necessitate some thought about what the "generic" header/body/footer styles should look like for Panels. @changecourse, would you like to set up a meeting sometime this week to discuss this?

src/webextension/list/components/item-list.css
```diff
@@ -19,3 +19,21 @@
   border-bottom: 1px solid #d7d7db;
   background-color: #ffffff;
 }
+
+.item:hover {
+  background-color: #ededf0;
+}
```
jimporter: I'm going to have to think a bit more about the changes in this file. The style rules for the item lists are where they are now to minimize momentary flashes of color when opening the item details view in the doorhanger. These changes may regress that, but I need to sit down and go through all the behaviors carefully to be sure I'm getting this right before I tell you to do anything different. :)

nickbrandt:
@jimporter Pushed some updates to this PR based on our discussions yesterday. Let me know if there are any other adjustments you are wanting to make (such as in regard to hover states comment above). I created a separate issue to update the height of the panels as well as refine the entry details now that we are updating the default header styling.

src/webextension/list/manage/containers/all-items.css
```diff
@@ -2,20 +2,6 @@
...
-  background-color: #ebF5ff;
-  border-bottom-color: #99bff2;
-  box-shadow: 0 -1px 0 #99bff2;
-}
-
-.all-items > .item:first-child[data-selected] {
-  border-top-color: #99bff2;
-}
-
 .all-items:focus > .item[data-selected] {
```
jimporter: Let's move this to `item-list.css` too. (This means we can delete this file entirely.)

src/webextension/list/manage/containers/all-items.css
```diff
@@ -2,20 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-.all-items {
-  border-inline-end: 1px solid #d7d7db;
```
jimporter: We need this border to follow the visual spec, but it really belongs elsewhere. Specifically, it should go in `src/webextension/list/manage/components/app.css` in the definition for `.app-main > aside`. (This also fixes some wonkiness in the styles there, which is nice.)

src/webextension/widgets/scrolling-list.css
```diff
@@ -14,16 +14,6 @@
   outline-style: none;
 }
 
-.styled-item {
```
jimporter: We should keep all this stuff here. Instead, we should change how `styledItems` is set. Right now, it's set in `src/webextension/list/manage/containers/all-items.js`; I think it should go in `src/webextension/list/components/item-list.js` so that it applies to all instances of the item list.

That does mean these CSS rules will be unused for now, but I expect we'll want this the way it is when we add support for multiple origins in the editor.

src/webextension/list/manage/containers/all-items.css
nickbrandt: How can I target the container focus (which was .all-items) from within the `item-list.css` file? From within `item-list.js` I don't see so a way to add a class to target to the container.

jimporter: You need to pass `className={whatever}` to the `<ScrollingList/>`, and then do pretty much the same CSS as here.

nickbrandt: I tried adding a class that way but it doesn't seem to be bringing it in (I have also defined the style in `item-list.js`).

screen shot 2018-03-08 at 11 06 07 am

jimporter: Did you also remove the `className` from `all-items.js`? If you didn't, it'll override the `className` here (because it's getting pulled in from `{...props}`, which occurs later on this line).

Maybe I should establish a general rule that `{...props}` goes first to avoid this issue.

nickbrandt: 👍 Got it, thanks for your help!

src/webextension/widgets/scrolling-list.css
nickbrandt: I just want to verify this is the correct implementation of this change:
`src/webextension/list/manage/containers/all-items.js`
Screen Shot 2018-03-08 at 11.46.39 AM

`src/webextension/list/components/item-list.js`
screen shot 2018-03-08 at 11 46 03 am

jimporter: I'll have to look at the full diff, but the only thing that jumps out here is that `style.allItems` should be `styles.itemList`. Also, don't forget to update /src/webextension/list/**popup**/containers/all-items.js similarly.

nickbrandt: I'm unsure of what needs changed for the `src/webextension/list/popup/containers/all-items.js` file. Can you let me know what needs to be updated here when you have a moment?

jimporter: It might actually be ok as it is. I thought we had some special styles in there, but it looks like no... vOv

nickbrandt:
@jimporter I believe I covered all the changes if you don't mind doing another review when you have a chance.

jimporter:
@nickbrandt I'm doing my review now, but just as a small note, when you're ready for another review, you can go up to the "Reviewers" area at the top of this page and re-select me as reviewer. That *should* put this issue in my list of "stuff I need to review", which makes it easier for me to keep on top of this stuff without forgetting. :)

src/webextension/list/components/item-list.css
```diff
@@ -19,3 +19,25 @@
...
+
+.item[data-selected] {
+  background-color: #ebF5ff;
+  border-bottom-color: #99bff2;
+  box-shadow: 0 -1px 0 #99bff2;
+}
+
+.item:first-child[data-selected] {
+  border-top-color: #99bff2;
```
jimporter: As far as I can tell, this doesn't actually do anything, since the top border is 0px tall.

src/webextension/list/components/item-list.css
```diff
@@ -19,3 +19,25 @@
...
+}
+
+.item-list:focus > .item[data-selected] {
+  box-shadow: inset 0 1px 0 #0a84ff, inset 0 -1px 0 #0a84ff;
```
jimporter: I think this is actually going slightly against the spec. This gives an "inner" border on the top/bottom of the row, in addition to the regular border. The CSS that you have above for `.item[data-selected]` looks more-accurate (where you specify a `bottom-border-color` and a `box-shadow` to draw on top of the bottom border of the previous element).

I'm basing this on the spec [here](https://app.zeplin.io/project/59fb4868c5e7608593f76965/screen/5a85d57f54a506c53acf1d98), although it specifies a state the user can't (currently) get to. In the spec, the selected row and the focused row are different, but they must always be the same the way our code works right now. Still, it *looks* like the focus indicator is supposed to change the color of the borders between each row.

All that said, we *do* use an "inner" border to represent focus state in a lot of places in our UI (e.g. the buttons on the footer of the doorhanger), so doing it the way you have now would probably be consistent; we do usually use an inner border of 2px on *all* sides though.

src/webextension/widgets/panel.css
```diff
@@ -13,8 +13,9 @@
   display: flex;
   align-items: center;
   font-weight: 500;
-  height: 32px;
-  padding: 2px 5px;
+  background: #ededf0;
```
jimporter: This is a good change for consistency with the item list and the details, but there's one tiny thing left: in `src/webextension/list/popup/components/app.css`, you'll need to set the background color of the body to match this. This will ensure that the little arrow above the doorhanger is the right color.

src/webextension/widgets/panel.css
```diff
@@ -13,8 +13,9 @@
...
+  height: 54px;
+  padding: 0 12px;
```
jimporter: Is there any reason for specifying the `height` and `padding like this instead of this?

```css
height: 30px;
padding: 12px;
```

That seems a little cleaner to me, but I don't feel super-strongly about this.

src/webextension/widgets/scrolling-list.css
```diff
@@ -26,4 +26,5 @@
 .scrolling-list:focus > .styled-item[data-selected] {
   color: #ffffff;
   background-color: #0060df;
+  box-shadow: inset 0 1px 0 #0a84ff, inset 0 -1px 0 #0a84ff;
```
jimporter: I don't think we need this change (at least not yet). For the time being, my intention is that a `<ScrollingList/>` with styled items should look like the search engine list in `about:preferences#search`. (This implementation makes it easier to see the focus state, though.)

jimporter:
Just a note: I believe I've resolved the merge conflict from my quick-copy PR.

jimporter:
@changecourse Could you take a look at how the focus/hover states look for "verbose" items (when we show the quick-copy buttons)? Currently, they highlight the entire item, *including* the quick-copy buttons, which I'm not sure is right.

Sorry for not catching this earlier...

nickbrandt:
Recent changes update the hover and focus to only apply to the entry summary area instead of including the buttons as well.
focus
hover

src/webextension/list/components/item-list.css
```diff
@@ -20,23 +20,26 @@
   background-color: #ffffff;
 }
 
-.item:hover {
+.item:first-child {
+  border-top: 1px solid #d7d7db;
```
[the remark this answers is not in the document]
changecourse: Option 4 makes sense to me. Our design also has a double border sort of appearance for the first item in the list on the doorhanger anyways, to give more separation between the search bar and the entry list.

src/webextension/list/components/item-list.css
```diff
@@ -20,23 +20,26 @@
...
-  box-shadow: 0 -1px 0 #99bff2;
+  box-shadow: 0 -1px 0 #99bff2, 0 1px 0 #99bff2;
 }
 
-.item-list:focus > .item[data-selected] {
-  box-shadow: 0 -1px 0 #0a84ff;
-  border-bottom-color: #0a84ff;
+.item-list:focus > .item[data-selected] .item-summary {
+  box-shadow: 0 -1px 0 #0a84ff, 0 1px 0 #0a84ff;
+  position: relative;
```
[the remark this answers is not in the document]
nickbrandt: I'll update this to be on the definition you suggested.

jimporter:
Ok, I think all my concerns are addressed (aside from adding a unit test, which I'll do shortly).

jimporter:
I added tests.

devinreams:
Thanks, @jimporter @nickbrandt for wrapping this up together.

@m8ttyB this is now awaiting final PI review. 📋 

Then we can go merge a couple other things, too! 🎉

devinreams:
@m8ttyB Thanks 🙇 

@jimporter I'm gonna squash this down into a single commit on `master` instead of pulling in all 15. I'll then go post updates in the dependent PRs...

PR #623, opened 2018-03-07
Release 0.1.8

devinreams:
@mozilla-lockbox/product-integrity this is what the "production" release will include and is ready for full/regression testing before tomorrow morning's release.

The commits view (as to what is included here) may look a bit confusing because the only _true_ changes coming along are those starting from Feb 27, 2018 onwards. Everything before that _technically_ is already on production, just in a way that git was unaware of (see prior email about this branch excitement for more detail).

We've got just one last change that will go in, the release notes at #612, but I don't think those will affect regression testing here. 😄 It's all yours!

- [x] PI approval
- [x] merge #612


devinreams:
>@m8ttyB approved these changes 

Thank you! 😍 

We're ready to rock! 🎸

PR #624, opened 2018-03-07
Add quick-copy buttons to the doorhanger listview; resolves #520

### LINKED ISSUE #520: Provide a quick method of copy/paste when there are one or two results for a URI to help users log into their account quickly

jimporter:
I think this is ready for review.


devinreams:
Nice! Works as described in #520. 👍

For those following along at home...

## Invoke the doorhanger with or without search terms

screen shot 2018-03-08 at 10 41 07 am

## Invoke the doorhanger with a URL pre-filled

copy-paste-doorhanger

changecourse:
Also curious... because our copy action is set to a standard timer, I'm able to click between buttons and have multiple states of 'copied' even though it's truly only the last item that I've clicked as tied to my clipboard... this functional change may need to be considered as a separate issue (by clicking copy in one area, automatically time out all other areas with 'copied' still active)

jimporter:
> Also curious... because our copy action is set to a standard timer, I'm able to click between buttons and have multiple states of 'copied' even though it's truly only the last item that I've clicked as tied to my clipboard... this functional change may need to be considered as a separate issue (by clicking copy in one area, automatically time out all other areas with 'copied' still active)

Could you file an issue on this? It'd be a nice improvement, but also a fair amount of work.

jimporter:
@changecourse Thanks for the UI review. I think I've gotten everything looking correct now.

PR #630, opened 2018-03-08
Move python runtime for deploy stage out of 'on' requirement

devinreams:
I think I made a dumb mistake at #613 and "required" python in the wrong way. Right now master deploys are skipping the pages deploy because this is in "on" (check if its in the runtime vs just set the runtime).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,12 +52,12 @@ jobs:
     - stage: deploy
       if: branch = master
       <<: *node
+      python: '3.6'
       script: skip
       deploy:
         provider: pages
         on:
           branch: master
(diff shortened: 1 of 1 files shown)


devinreams:
>Couldn't push the build to github.com/mozilla-lockbox/lockbox-extension.git:gh-pages

The failures to deploy seems to coincide with an upgrade within `dpl` (Travis CI deploy scripts) that broke pages releases: https://github.com/travis-ci/travis-ci/issues/9312 😅

Attempting a work around here: https://github.com/mozilla-lockbox/lockbox-extension/pull/643

PR #553, opened 2018-02-20
Styling updates around entry list items

nickbrandt:
Partially fixes #475 - A few tasks left which will need dev assistance

Updates include:
-Active and focus states
-Increased padding to match mocks
-New chevron icon
-Updated aside background color
-Moved border between search and entries in order to apply active border color to top entry item
-Small copy change to empty state

PR #642, opened 2018-03-09
Add a top-border above the item list in the manager

jimporter:
Depends on #619 

This is based on #619, and just fixes a little thing I noticed during review, but which is out-of-scope for that PR. Since I didn't want to start causing scope creep there, here's a tiny PR to fix things. Only the last commit is relevant.

### DIFF
--- a/src/webextension/list/manage/components/app.css
+++ b/src/webextension/list/manage/components/app.css
@@ -56,3 +56,7 @@ body > main,
 .filter-bar {
   padding: 12px;
 }
+
+.all-items {
+  border-top: 1px solid #d7d7db;
+}
--- a/src/webextension/list/manage/components/app.js
+++ b/src/webextension/list/manage/components/app.js
@@ -45,7 +45,7 @@ export default class App extends React.Component {
                                 this._filterField = element;
                               }}/>
                 </Toolbar>
-                <AllItems/>
+                <AllItems className={styles.allItems}/>
               </aside>
               <article>
                 <CurrentSelection/>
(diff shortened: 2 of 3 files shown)


jimporter:
Note: this has r+ and ui-r+; just waiting on the parent PR to merge before I merge this.

nickbrandt:
@jimporter We had removed this border in #553 until we redid the filter area. Can we instead apply the border to the first entry list item, as that would make it easier to change the border color for that entry on active state? I had actually implemented that change back in #657 , and you can see there how it's currently working. I can strip that out now knowing there is this issue in place already, but wanted to see if we can add the border in that manner instead.

jimporter:
@nickbrandt See my comments on your PR for this. However, the tl;dr is that I think we probably want this border to *not* be attached to the first item.

devinreams:
@jimporter @nickbrandt Now that #619 is merged this is ready for rebase (see conflicts) and merge, assuming you all are.

jimporter:
@changecourse Just wanted to get one last peek at this to make sure this is what you wanted for the "double border".

jimporter:
@changecourse That does mean there's *no dividing line* between the search box and the item list when scrolled down past the first item. Is that what you want?

changecourse:
@jimporter agh, right... yeah, not my intent either.

What about the option #3 then, as referenced from your previous comments on this... 

> 3. Instead of using a bottom-border on the header, give it a box-shadow on the bottom with a higher z-index than the list, and give focused/selected items an even higher z-index; this will look slightly off when a focused/selected item is partially scrolled out of view though

Could this approach work for both the doorhanger and full tab management view?

jimporter:
@changecourse Ok, I think I actually have a "perfect" solution, but it's somewhat messy and will require moving a few things around. I'll do this in the PR I'm working on to remove the main toolbar from the management tab (since I'm already making some changes there that make this much easier).

PR #646, opened 2018-03-09
Add a fallback for the no-results case in the doorhanger; resolves #515

### LINKED ISSUE #515: Provide a full list of options when there are no-results for the current URI to mitigate confusion for why there aren't relevant results applicable to that URI

jimporter:
This is still WIP and needs tests (plus rebasing on top of a few other PRs), but it basically works. Here's how I've handled the fallback cases:

No filter + empty DB => "No results"[1]
User filter + empty DB => "No results"[1]
User filter + no matches => "No results"
Auto filter + empty DB => "No results"[1]
Auto filter + no matches => Fallback to show everything

Note: the stuff marked [1] could be adjusted to have some special "get started" text if we want, or an explanation that the DB is empty (we do this in the management UI). I'll let @changecourse chime in on this.


jimporter:
Just a note for you @changecourse, I flagged you for review, but I'm just looking for some feedback on the UX right now. I'd like to wait to fix up the visuals until some of our other work has landed. That'll make it a lot easier to get something nice-looking.

jimporter:
@changecourse Ok, I've rebased on top of @nickbrandt's work on the doorhanger now, so feel free to look at all the visuals. One thing I'm missing though it the little "i" in a circle on the "no results" banner. If you could post the SVG for that, I'll add it in. I also had to guess on the styles for the banner, since Zeplin wasn't letting me inspect it. Hopefully it's right.

changecourse:
Additionally, I may have ran into a potential issue around handling detection of stories... If I only have one or two entries in my Lockbox total, and I head to a website that DOES NOT match via URI detection, I'm still seeing the quick copy buttons for the entries in the doorhanger... we should discuss as a group whether or not this is problematic... 

As of now, I'm ok with this staying the way it is (after all, we do have the real estate) but I figured I'd call it out.

jimporter:
@changecourse That's intentional. We had agreed that the quick copy buttons show up when there are <=2 results. The initial comment in #520 alludes to this.

changecourse:
Sounds good, thanks for the clarification, makes sense.

devinreams:
Since #619 merged I believe this branch got a little stale and now has conflicts. ❌ 

And @changecourse [requested](https://github.com/mozilla-lockbox/lockbox-extension/pull/646#pullrequestreview-103125326) some UI revisions. ❌ 

Once those are addressed I believe this branch will be ready for final UI review, code review, and PI. 📋

devinreams:

This assumes the final UI work, reviews, and final testing can reasonably be wrapped up between tomorrow and Wednesday before regression testing. Anyone can correct me if wrong...

jimporter:
Hmm, the build is busted. Let's figure out why...

jimporter:
Ok, tests should be working. This can land once the build is green.

devinreams:
Thanks @jimporter. Sorry I missed the breakage ❌. I think I'm used to seeing red buttons and just pressing them as admin. That's not... ideal. 😔

m8ttyB:
@devinreams I prefer to keep new feature testing outside of the scope of end-to-end regression testing if you want in-depth testing done. If the risk is such that we don't need PI to take a close look at this work I can stick to happy path verification of the feature.

devinreams:
>I prefer to keep new feature testing outside of the scope of end-to-end regression testing if you want in-depth testing done.

Makes sense. Then this is definitely a candidate for that. I was meaning more a "batch up your work" all at once, as opposed to look at this once, and then pick it all up and look at it again. Sorry for being confusing..

m8ttyB:
>  Sorry for being confusing..
Ah, I see my confusion. I thought you meant merge this to master, yeah, this is landing later in the day than is ideal but I can include it in the test cycle before beginning regression testing.

m8ttyB:
Just so I'm clear on the behavior of this feature, I have a few questions.

When the URI auto-filters and there is no entry in the DB, the language "No entries found for the current website" is present in the door hanger is present and all entries are shown.

screen shot 2018-03-21 at 2 39 55 pm

When the user sets the search filter manually and there is no entry in the DB we don't display the No entries found for the current website" language?

screen shot 2018-03-21 at 2 40 12 pm

jimporter:
@m8ttyB That is correct.

PR #655, opened 2018-03-13
Changing height of doorhanger to be consistent across instances

### LINKED ISSUE #620: DH: Update height of doorhanger to be consistent throughout usage
We are updating the height of the doorhanger to be consistent in it's use in Sign-in(Unlock), Entry List & Entry Details. The new height value is to be **360px**, which will increase the size for the Unlock screen and decrease it a little (ability to show 3.5 entries visible before scrolling) for the entry list & detail.

- [x] Update Sign-in (Unlock) height
- [x] Update Entry List & Details height

nickbrandt:
Resolves #620 

- Changed height to be 360px across instances (unlock, entry list, entry details).
- Updated some unlock (sign-in) styles to accommodate the change.


devinreams:
@nickbrandt Now that #619 is merge, there is a conflict in the `popup/components/app.css` file you'll want to manually review and see what all should stay/go/change.

devinreams:

This assumes the a final 👍 review from @jimporter within the next day or so...

src/webextension/unlock/components/app.css
```diff
@@ -13,9 +13,13 @@ body {
   -moz-user-select: none;
 }
 
+.unlock {
+  height: 360px;
```
jimporter: Let's move this up to the `body` element, where `width` is defined.

src/webextension/list/popup/components/app.css
```diff
@@ -14,5 +14,5 @@ body {
 
 .app {
   width: 280px;
-  height: 400px;
+  height: 360px;
```
jimporter: Can we also move this up to the `body` (`width` and `height`) like in the unlock view? Then we can probably remove this selector entirely (and its reference in `app.js`).

src/webextension/unlock/components/app.css
nickbrandt: Defining the height on the body instead of either the <main /> or <article /> will not create the desired effect (unless we also define those as being height: 100%). If that is preferred I can make the change, the same applies to the popup.

jimporter: Hmm, that might be a bug in our `<Panel/>` CSS. I'll take a look.

nickbrandt: I don't think it's a bug, simply that the footer of the panel is fixed to the bottom of the `<article />`, which in both instances has a wrapper which isn't set to 100% height.

jimporter: Oh wait, I understand. Let's do something similar to the management UI:

https://github.com/mozilla-lockbox/lockbox-extension/blob/b640724d02b0dac86234c00099ace8e53507bd46/src/webextension/list/manage/components/app.css#L18-L21

I think that'll make it extra-clear that the dimensions should apply to the size of the entire popup window.

src/webextension/list/popup/components/app.css
```diff
@@ -5,14 +5,16 @@
...
   font: caption;
   font-size: 13px;
   -moz-user-select: none;
 }
 
+body > main,
 .app {
-  width: 280px;
-  height: 400px;
+	height: 100%;
```
devinreams: maybe I'm crazy but should this backdent one? if so: no need to fix now, but if someone notices it in another round of revisions.

jimporter: Looks like a tab character snuck in.

nickbrandt: Doh...my tab size was set back at 4 somehow...my bad

devinreams: No problemo. 1fe3cc924ef083ec1fd85543d56699002a2ca97a

Issue #627, opened 2018-03-08
URI prefill only search by origin


sandysage:
When I land on my Gmail account, I want to only see the relevant results for https://mail.google.com/, and not all my google accounts and properties.

linuxwolf:
@sandysage 

I'm not sure this makes sense, and contradicts the notes/gist references in #527.  This is particularly true for sites that have separate sub-domains for "desktop" (almost always "www."; e.g., "www.facebook.com") versus "mobile" (e.g., "mobile.facebook.com").

With the above, the results may be "no entries" when the user's perception of their entries is that there is (at least) one.

linuxwolf:
I think I now see the intent, which is to filter the doorhanger by only site name (origin) when it is first opened.

jimporter:
@linuxwolf Yeah, I should probably go in and update the original comment with some more details. I think I alluded to it elsewhere, but the idea is that we *shouldn't* match on, say, the username field.

PR #660, opened 2018-03-14
[WIP] Parse URLs when filtering

jimporter:
This is still very WIP, but it should give an idea of how I plan to pass the data around. The general idea is that when the popup opens, we fill in the protocol, host, and port (if the port is default, it's implicitly defined by the protocol), which then gets passed to the filter code.

The filter code rebuilds a URL object and tries to match it to URL objects for the origins of each item. Origins are allowed to omit their protocol, in which case we treat them as HTTPS. If the search query omits *its* protocol (this will only happen when the user is typing in the filter themselves), we just do a string comparison. This way, you only have to type part of a URL (e.g. "goo" for "https://www.google.com") to filter the results down when you're searching manually.

Up next, I plan to make the matching on valid URLs work closer to our plan.

### DIFF
--- a/src/webextension/list/filter.js
+++ b/src/webextension/list/filter.js
@@ -7,16 +7,42 @@ export function parseFilterString(filter) {
   if (!filter) {
     return [];
   }
-  return filter.split(/\s+/).map((i) => i.toLocaleLowerCase());
+  return filter.split(/\s+/);
+}
+
+function tryMakeURL(string, defaultProtocol = null) {
+  let url = null;
+  try {
+    url = new URL(string);
+  } catch (e) {
+    if (defaultProtocol) {
+      try {
+        url = new URL(defaultProtocol + "//" + string);
+      } catch (e) {}
+    }
+  }
+  return {url, string};
 }
 
 function match(filterElement, value) {
-  return value.includes(filterElement);
+  return value.toLocaleLowerCase().includes(filterElement.toLocaleLowerCase());
 }
 
-function matchAny(filterElement, values) {
+function matchURL(filterElement, value) {
+  // If the user is in the middle of typing a URL or entered something not
+  // URL-like in the origin field, just to a string comparison. This is safe for
+  // strictly matching auto-filled URLs since they're always full hosts with
+  // protocol (and possibly port).
+  if (!filterElement.url || !value.url) {
+    return match(filterElement.string, value.string);
+  }
+  return (filterElement.url.protocol === value.url.protocol &&
+          filterElement.url.host === value.url.host);
+}
+
+function matchAny(filterElement, values, matcher = match) {
   for (let i of values) {
-    if (match(filterElement, i)) {
+    if (matcher(filterElement, i)) {
       return true;
     }
   }
@@ -24,12 +50,13 @@ function matchAny(filterElement, values) {
 }
 
 export function filterItem(filter, item) {
-  const title = item.title.toLocaleLowerCase();
-  const username = item.username.toLocaleLowerCase();
-  const origins = item.origins.map((i) => i.toLocaleLowerCase());
+  const title = item.title;
+  const username = item.username;
+  const origins = item.origins.map((i) => tryMakeURL(i, "https:"));
 
   for (let i of filter) {
-    if (!match(i, title) && !match(i, username) && !matchAny(i, origins)) {
+    if (!match(i, title) && !match(i, username) &&
+        !matchAny(tryMakeURL(i), origins, matchURL)) {
       return false;
     }
   }
--- a/src/webextension/list/popup/index.js
+++ b/src/webextension/list/popup/index.js
@@ -26,7 +26,7 @@ browser.tabs.query({ active: true, currentWindow: true }).then((result) => {
   const validProtocols = ["http:", "https:"];
   const url = new URL(result[0].url);
   if (validProtocols.includes(url.protocol)) {
-    store.dispatch(filterItems(url.hostname));
+    store.dispatch(filterItems(url.protocol + "//" + url.host));
   }
 });
 


src/webextension/list/filter.js
```diff
@@ -7,29 +7,56 @@ export function parseFilterString(filter) {
   if (!filter) {
     return [];
   }
-  return filter.split(/\s+/).map((i) => i.toLocaleLowerCase());
+  return filter.split(/\s+/);
+}
+
+function tryMakeURL(string, defaultProtocol = null) {
```
linuxwolf: For what the user types, or the extension grabs from the activeTab (let's call this `target`), using exactly what is provided is appropriate.  For what we get from a saved entry in Lockbox (even if the entry had "http:") (let's call this `allowed`), I am far less sure.

Matching on "http:" carries a lot of risk.  It's ok to for "http://localhost' (or "http://127.0.0.1", or "http://[::1]"); this is considered a secure origin (at least for now).  For other hosts/ports, I worry we are potentially exposing users to exfiltration.  Also, matching on other origins carries risk, too (e.g., "javascript:").

This is probably a long way of saying the data massaging here needs to be more robust; my personal preference is sooner than later, but it'll need to happen.

\[I do see the valid protocol constraints, but they are far from the rest of this constraints logic]

src/webextension/list/filter.js
```diff
@@ -7,29 +7,56 @@ export function parseFilterString(filter) {
...
   for (let i of values) {
-    if (match(filterElement, i)) {
+    if (matcher(filterElement, i)) {
       return true;
     }
   }
   return false;
 }
 
 export function filterItem(filter, item) {
```
linuxwolf: Looking at these changes, my sense is filtering needs to get more differentiating -- instead of relying on a string that is (potentially mis)interpreted in a few ways, it's an object that better outlines what the filter intent is.

src/webextension/list/popup/index.js
```diff
@@ -26,7 +26,7 @@ browser.tabs.query({ active: true, currentWindow: true }).then((result) => {
   const validProtocols = ["http:", "https:"];
   const url = new URL(result[0].url);
   if (validProtocols.includes(url.protocol)) {
-    store.dispatch(filterItems(url.hostname));
+    store.dispatch(filterItems(url.protocol + "//" + url.host));
```
linuxwolf: I would have expected `url` to be passed in directly here, rather than breaking it up.  This feels like the constraints are spread around the codebase; that will make it harder for us to understand the implications and limitations.

jimporter: We can't (well, we *shouldn't*) pass the URL object, since that goes right into the Redux state, and Redux is supposed to be just JSON. We could make a new Redux action creator, but it'd do pretty much exactly what this does before getting stored in Redux.

src/webextension/list/filter.js
jimporter: I'm not sure what this means. What *part* of filtering needs to be converted from a string to an object? The visible representation of a filter should be a string, since it's what goes into the text field, and one of the main reasons for having filters in the first place is as something a user could type.

If you're suggesting that the automatically-generated filter should only find results in the `origin` field of the item, I agree and that's #627. I'd like it to be something like Google's mini-format for search queries, e.g. `url:https://www.example.com` (there'd be other prefixes too, like `title:`, `username:`, and maybe one day, `tag:`).

src/webextension/list/filter.js
jimporter: Where would we want this validation to happen? If we want to be strict about what's allowed in the datastore, should we just throw an exception when trying to save an entry if the user specifies a "bad" origin? At the moment, this code is assuming that the origin could be *literally anything*, since we have no constraints on what a user enters into the datastore. However, we could mandate that origins take a particular form, and possibly even normalize them when saving them (e.g. `www.google.com` is normalized to `https://www.google.com`). I'd actually prefer this to what I have now, since it's one less place where I have to make guesses about what the user meant.

For HTTP, how do we balance the fact that some sites just don't use HTTPS (and users would like to log into them) with the potential for some kind of MITM attack? Note that the code here only applies to the doorhanger, which is in chrome, not content. When we have a matcher used by (semi-) autofill, we may well choose to be even more restrictive. For example, I'd be ok with a plan where we say, "Autofill only works on HTTPS, but you can get pre-filtered results in the doorhanger even on HTTP".

I also think it would make sense to provide more restrictions here (e.g. constraining protocols), but we'd probably *also* have the constraints where they are now (or in a Redux action creator) since I don't think we should even fill in the search box on URLs with "bad" protocols. That said, a lot of this stuff will probably move to a common file somewhere once it's fleshed out a bit better.

jimporter:
@linuxwolf Good call on using `psl`. I figured there had to be something like that, but I just wanted to sketch out a simple implementation of the URL checking, focusing on how the information passes from the index on page load, through the Redux store, and to the filtered output in our React UI. Here's the stuff I've tried to address in this WIP:

1. When opening the doorhanger, we initialize the contents of the search box to be something that a) has all the info we need, and b) makes sense to the user (i.e. it should be possible for a user to perform a manual search exactly like the automatic search). To me, this means "initialize the search field to `protocol://host[:port]`", since that's something a user could enter. In particular, I'd like to avoid any internal "magic" (e.g. a hidden flag that tells us the filter is autogenerated and should be handled differently), since I think that has a lot of potential for bugs and/or user confusion.
2. If the search field doesn't have a protocol, we know that it must be user-entered, so we do a regular ol' substring match. This is so that as the user types, the list of displayed entries gradually gets smaller. If we did a strict URL match, the user would see 0 results until they typed the whole URL. That'd suck, since you wouldn't be able to just type "goo" to match "https://www.google.com".
3. Since our Redux representation of the filter is just a string (that's what we'd get if a user just typed it in), we need to parse that just prior to performing the filter process into some kind of fancy object. That lets us perform a richer comparison than just string equality. For now, that just means using the `URL` constructor, but later, it could mean using `psl` or a combination of the two.
4. Since we have no validation on what gets stored in the origin field in `lockbox-datastore`, I made the temporary decision that we'd just do our best to parse it as a URL (defaulting to HTTPS if there's no protocol). However, we could add some requirements about what can be stored in the origin field, which would eliminate the need for this guesswork.

src/webextension/list/filter.js
linuxwolf: > Where would we want this validation to happen? If we want to be strict about what's allowed in the datastore, should we just throw an exception when trying to save an entry if the user specifies a "bad" origin? At the moment, this code is assuming that the origin could be *literally anything*, since we have no constraints on what a user enters into the datastore. However, we could mandate that origins take a particular form, and possibly even normalize them when saving them (e.g. `www.google.com` is normalized to `https://www.google.com`). I'd actually prefer this to what I have now, since it's one less place where I have to make guesses about what the user meant.

The datastore should have been more restrictive, and we can (should) make it more restrictive now (ish).  We still have to deal with existing data, though.

> For HTTP, how do we balance the fact that some sites just don't use HTTPS (and users would like to log into them) with the potential for some kind of MITM attack? Note that the code here only applies to the doorhanger, which is in chrome, not content. When we have a matcher used by (semi-) autofill, we may well choose to be even more restrictive. For example, I'd be ok with a plan where we say, "Autofill only works on HTTPS, but you can get pre-filtered results in the doorhanger even on HTTP".

Since _this_ code only applies to doorhanger's filter, I suppose _**for now**_ it's ok to match on `http:` also.  Howver, it is **MUST NOT** to extend that to our upcoming fill work, unless security-review finds it acceptable.

src/webextension/list/filter.js
```diff
@@ -7,29 +7,56 @@ export function parseFilterString(filter) {
...
+function matchURL(filterElement, value) {
+  // If the user is in the middle of typing a URL or entered something not
+  // URL-like in the origin field, just to a string comparison. This is safe for
+  // strictly matching auto-filled URLs since they're always full hosts with
+  // protocol (and possibly port).
+  if (!filterElement.url || !value.url) {
+    return match(filterElement.string, value.string);
+  }
+  return (filterElement.url.protocol === value.url.protocol &&
+          filterElement.url.host === value.url.host);
```
linuxwolf: looking at this more, I'm less and less comfortable with the missing port.

src/webextension/list/popup/index.js
linuxwolf: That is a reasonable argument against passing the URL directly.

However, this still needs to be `hostname` and not `host`.  It needs to keep the optional port, or it risks false positives.  For the vast majority of people for the vast majority of their entries, `hostname` and `host` are identical, so I think false negatives by real users will be rare.

I can see an argument for treating the lack of an explicit port as matching any port, but I'd like to get sign off from security-review.

src/webextension/list/filter.js
jimporter: `URL.host` includes the port if it's not the default port for that protocol. So for `https://www.google.com`, `URL.host` is `www.google.com`, but for `https://www.google.com:8000`, `URL.host` is `www.google.com:8000`. I plan to make this a lot more explicit when I build some common functions to use here (i.e. I'll write documentation).

linuxwolf: I apologize for that; I keep having them swapped in my head.

jimporter: No worries; this is why we need documentation (and instinctively, `hostname` is longer than `host`, so the value seems like it should be longer too!).

PR #665, opened 2018-03-16
Create a classNames() function to handle building className props more easily

jimporter:
This is effectively a simplified version of the npm `classnames` package. This commit also moves `{...props}` to the beginning of our list of applied props so that it's no longer possible to accidentally overwrite internal props.

This is a fairly big diff, but it's pretty simple, and should make it easier to apply our CSS class names. (I'm doing this now because I've been finding it an issue while working on the breadcrumbs PR.)

### DIFF
--- a/src/webextension/common.js
+++ b/src/webextension/common.js
@@ -10,3 +10,7 @@ export function makeItemSummary(item) {
     username: item.entry.username,
   };
 }
+
+export function classNames(classNames) {
+  return classNames.filter((i) => i).join(" ");
+}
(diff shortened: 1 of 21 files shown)


jimporter:
@nickbrandt I mainly flagged you for review to give you a heads-up that I'm adding this and to make sure it's something that would be an improvement for the occasions you have to dig into our React code. :)

devinreams:

This assumes a 👍 from PI review within the next day or so..

PR #666, opened 2018-03-16
Improve DH integration tests and reorganize test suite

### LINKED ISSUE #663: Improve the DH integration tests
I'd like to see the door hanger test improved to include:
* tests reorganized into functional suites that match TestRail (scoped to include the non-DH tests)
* logging into FxA
* when creating the entry, putting content into each of the input fields
* instead of verifying the entry fields are blank, update the test to verify the name is correct

A future task will be to improve the test to validate the remainder of the entries data data fields are correct.

m8ttyB:
Resolves #663 

* A slight update to the existing door hanger test.
* Updates `new_entry()` to `create_new_entry()` which can take args.
* Begins to move away from the `munged_class_name()` helper and uses CSS selectors.
* Moves the remaining test cases into suites that match the canonical suites in TestRail.

### DIFF
--- a/test/integration/pages/home.py
+++ b/test/integration/pages/home.py
@@ -27,8 +27,6 @@ class Home(Base):
                           'section menu '
                           'button.{}:nth-child(1)'.format(
                             munged_class_name('button')))
-    _save_entry_locator = (By.CSS_SELECTOR, 'article div form menu '
-                           'button.{}'.format(munged_class_name('button')))
     _sign_in_locator = (By.CSS_SELECTOR, '.{} .{}'.format(
                         munged_class_name('link'),
                         munged_class_name('puffy-size')))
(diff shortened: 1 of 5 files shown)


test/integration/pages/home.py
```diff
@@ -50,11 +48,18 @@ def lockie(self):
...
+        """Create and save a new entry."""
         current_entries = len(self.entries)
         self.find_element(*self._new_entry_locator).click()
-        self.find_element(*self._save_entry_locator).click()
+        entry = self.entries[0].click()
+        entry.set_site_name(site_name)
+        entry.set_website(url)
+        entry.set_username(username)
+        entry.set_password(password)
+        entry.set_note(note)
```
jrbenny35: Is there a reason we would only set 1 of these?

m8ttyB: Do mean can a user have multiple notes per entry? At this time, no.

jrbenny35: No, I am asking if we would only use ```set_username``` for example, or would it be all of them every time?

m8ttyB: I follow now, thanks. I think for this first iteration I was thinking yes. The next round of work I want to move things around a bit more and could see this method returning an Entry object and the individual test touches the setter methods. Do you see this as blocking or can this happen in a separate pr?

jrbenny35: Not it's not blocking. I think we could iterate on this though. Maybe we don't need so many ```set_``` methods.

m8ttyB: Agreed. I felt unclean doing so.

devinreams:

This assumes a 👍 review from @jimporter within the day or so.

PR #669, opened 2018-03-16
Replace the main toolbar with breadcrumbs

### LINKED ISSUE #444: IA Updates: Header Bar & Creating an Entry
Improving the relationship between the "Create Entry" button and the entry list, based on past usability concerns (from Alpha) and feedback from the team during all-hands. 

(EDIT: cleaned up past conversation and feedback on outdated designs)

## Visual Design

**InVision**
https://mozilla.invisionapp.com/share/CTEFRXMZ7#/262163357_1-1-1_Lockbox_Home_-_View_Entry

**Zeplin**
https://zpl.io/begk4BD

**Tasks (may need to be broken up into separate issues)**
- [x] Removal of the home button
- [x] Navigating to home will now be accomplished via breadcrumb navigation, only available when at least one level deep (such as within the entry detail, see designs) Breadcrumbs should not be displayed at the root level (Home)
- [x] Revision of the 'entry button' to match the revised designs (now as a '+' alongside the search)
- [x] Removal of the header background bar (sidebar is now flush with the top, account and FAQ links float in the Home content area
- [x] Move the feedback link to the base of the entries sidebar, as a fixed callout

jimporter:
Resolves #444

Remaining work:
* [x] Unit tests for the new components to handle the "no entries" border fix
* [x] Unit tests for the breadcrumbs
* [x] Fix integration tests (assuming they're broken)

---

Old, irrelevant comments here:

@changecourse Could you take a look at how the borders for the entry list look? I think I got things working now. There are two caveats:

1. There's a border between the search box and the "no entries" callout; the spec says there should be no border. I'll fix this later.
2. When the entry list has enough items to scroll, the grey borders at the top and bottom overlap the scrollbar by 1px. I don't think I can do much about this without sacrificing other, more-noticeable aspects of the visuals. (If we ever add overlay scrollbars, we could probably fix this then.)


devinreams:
Quick WIP preview of this branch:

screen shot 2018-03-19 at 9 30 49 am

changecourse:
Borders look good! @jimporter

jimporter:
@devinreams I notice there aren't any scrollbars in your screenshot. Is everything working ok with scrolling on your end? I see scrollbars in my local build, so maybe Firefox has overlay scrollbars on macOS now?

changecourse:
@jimporter correct, Firefox on Mac includes overlay scrollbars by default (which why I was previously confused about you previous comments regarding the need to implement them)

jimporter:
@changecourse Visually, this should be done, if you'd like to take a look at everything.

src/webextension/list/manage/components/home-breadcrumbs.js
```diff
@@ -0,0 +1,41 @@
...
+import PropTypes from "prop-types";
+import React from "react";
+
+import Breadcrumbs, { Crumb } from "../../../widgets/breadcrumbs";
+import { NEW_ITEM_ID } from "../../common";
+
+export default function HomeBreadcrumbs({itemId, itemTitle, onClickHome}) {
+  if (itemId) {
+    const trimmedTitle = itemTitle.trim();
+    const titleId = `breadcrumbs-item-${itemId === NEW_ITEM_ID ? "new" : ""}`;
```
linuxwolf: This results in a trailing '-', and causing bad display.  I think this ought to be:
```
`breadcrumbs-item${itemId === NEW_ITEM_ID ? "-new" : ""}`
```

As-is, this renders as:

image

linuxwolf:
The breadcrumbs display jumps around a tiny bit if the entry's title contains some (or all) emoji.  On my computer (MacOS 10.13.3 + Firefox Nightly 61.0a1 (2018-04-02) (64-bit)) the content height is 18.5px for Latin-1 characters and 20px for emoji.  I don't know how much it's worth tracing through right now, though.  It's pretty subtle.

sandysage:
> The breadcrumbs display jumps around a tiny bit if the entry's title contains some (or all) emoji.

That seems reasonable to leave 'as-is' given that we don't have any insight as to expected emoji usage and needed support.

jimporter:
@linuxwolf This is finally ready for a final review.

PR #680, opened 2018-03-20
Hide "clear" button on filter when there is no input

### LINKED ISSUE #622: Only display 'clear' icon if search bar has something in it
Clear icon in search bar should not be shown unless there is something inside of the search field.

devinreams:
Fixes #622 

This changes the styles and checks the state so that the clear ❌ button does not show on the search/filter input until there is actually text in the field.

hide-search-button

@jimporter I think this is the right way to check both for the input value _and_ inherit the entire form's `disabled` prop...

### DIFF
--- a/src/webextension/widgets/filter-input.js
+++ b/src/webextension/widgets/filter-input.js
@@ -67,7 +67,7 @@ export default class FilterInput extends React.Component {
                ref={(element) => this.inputElement = element}/>
         <Localized id="filter-input-clear">
           <button type="button" title="cLEAr"
-                  disabled={disabled ? disabled : !this.state.value}
+                  disabled={disabled || !this.state.value}
                   onClick={() => this.updateValue("")}/>
         </Localized>
       </div>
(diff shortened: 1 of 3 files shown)


src/webextension/widgets/filter-input.js
```diff
@@ -66,7 +66,8 @@ export default class FilterInput extends React.Component {
                onChange={(e) => this.updateValue(e.target.value)}
                ref={(element) => this.inputElement = element}/>
         <Localized id="filter-input-clear">
-          <button type="button" title="cLEAr" disabled={disabled}
+          <button type="button" title="cLEAr"
+                  disabled={disabled ? disabled : !this.state.value}
```
jimporter: `disabled || !this.state.value` is probably simpler. I'm not sure how this would interact with a11y stuff though...

devinreams: >`disabled || !this.state.value` is probably simpler

Oh, duh.. thanks.

>I'm not sure how disabling this (rather than removing the element entirely) would interact with a11y stuff though

Fair. Per the spec, the [disabled attribute](https://www.w3schools.com/tags/att_button_disabled.asp) means the button "is unusable and un-clickable" so feels like its going to be safely ignored anyway. I see that reflected in the fact that 'tabbing' does not try to stop on this hidden element, it goes straight from the input to the item list.

PR #687, opened 2018-04-12
Update to the latest version of fluent

jimporter:
Note: We had to switch to the "compat" versions of the fluent libs, since they use some bleeding-edge JS features that made babel/webpack unhappy. We should figure this out and come up with a better solution at some point!


jimporter:
@stasm I had to switch to using the "compat" forms of `fluent` and `fluent-react` in order to make Babel/Webpack happy. Are there any examples of using Fluent with Babel/Webpack that I should be following instead, or is the code in this PR how I should be doing things?

stasm:
> @stasm I had to switch to using the "compat" forms of fluent and fluent-react in order to make Babel/Webpack happy. Are there any examples of using Fluent with Babel/Webpack that I should be following instead, or is the code in this PR how I should be doing things?

Thanks for asking. This PR is exactly what I would recommend—for web projects. I do wonder if it's needed in the context of a web extension. Do you need to support versions of Firefox older than 57?

I hope it wasn't too much trouble. Were you using `babel-polyfill` before this PR?

jimporter:
> I do wonder if it's needed in the context of a web extension. Do you need to support versions of Firefox older than 57?

No. In fact, we could probably increase the minimum version of Firefox to the latest nightly if we wanted.

> Were you using `babel-polyfill` before this PR?

I think we were using it at one point and then (mostly) removed it. If we could avoid `babel-polyfill`, that would be nice.

src/webextension/firstrun/components/app.js
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { Localized } from "fluent-react";
+import { Localized } from "fluent-react/compat";
```
linuxwolf: \\-:

.babelrc
```diff
@@ -4,6 +4,7 @@
     ["env", {"targets": {"browsers": ["firefox >= 56"]}}]
   ],
   "plugins": [
+    "transform-async-generator-functions",
```
linuxwolf: /-:

jimporter:
> Maybe you don't need babel-plugin-transform-async-generator-functions then?

Yeah, that was an accidental change that slipped in when I was trying stuff out. I'll see if I can get everything working a little more cleanly. Thanks!

stasm:
Great, I hope it works :) Let me know if you run into any troubles and thank you for blazing the path. I'd like to document this for other users of `fluent-react` who don't need to support older browsers.

jimporter:
@stasm Ok, I *finally* figured it all out! I'll write up a little guide on Monday, but I'll also be filing a bug against `fluent-react`, since there's an issue there that prevents consumers of the lib from doing everything nicely.

jimporter:
@stasm Filed the issue as https://github.com/projectfluent/fluent.js/issues/192.

PR #696, opened 2018-05-03
[WIP] Update Fluent (second take)

jimporter:
Supersedes #687


jimporter:
@linuxwolf Ok, it looks like the only thing broken now is that node-jose (included via lockbox-datastore) uses a vulnerable version of base64url.

linuxwolf:
@jimporter   Updates for that are coming!

PR #738, opened 2018-07-18
Add contributing notices to clarify expectations

devinreams:


pdehaan:
Should we just archive the repo?
Code should all still be available, but possibly a bit more obvious it’s not in active development.

devinreams:
@pdehaan Yeah that's a great idea. I decided not to _quite_ yet for a few reasons:

1. we may still pick up and do some work here in the coming weeks/month
2. I'd like to make sure issues can be opened and dependencies can be updated for a bit still
3. we may do another release to upgrade or notify users in Firefox itself

Consider this a half-step towards potential archival...

